PR #4, opened 2021-07-23
Place an (invisible) cursor at the beginning of the active list item

mlang:
This change is to support screen readers which use the cursor location
to indicate the focus to the user.

Brick.putCursor is unreleased, so grab the latest version from git via extra-deps.


hasufell:
Merged, thanks!

hasufell:
This is now in the current 0.1.16.1 release.

mlang:
I am already using it, thanks for a very fast turnaround.
I believe this is the fastest I ever got an accessibility issue fixed
and shipped ever!

PR #694, opened 2022-11-22
Fix armv7 builds

hasufell:


hasufell:
@newhoggy cabal-cache seems to do nothing, but doesn't error either: https://github.com/haskell/ghcup-hs/actions/runs/3552070582/jobs/5967056739#step:5:17079

I checked the bucket, it's empty. `CABAL_DIR` is set to a custom location.

hasufell:
It seems only the self hosted runner is able to use cabal-cache with backblaze S3. I wonder if github runners have this blocked?

hasufell:
So with debug turned on I get

```
2022-11-27T03:01:48.7246401Z [Error] [tid: ThreadId 9][ServiceError] {
2022-11-27T03:01:48.7246691Z   service    = S3
2022-11-27T03:01:48.7246928Z   status     = 404 
2022-11-27T03:01:48.7247171Z   code       = 
2022-11-27T03:01:48.7247423Z   message    = Nothing
2022-11-27T03:01:48.7247784Z   request-id = Just da324536ddfccbd2
```

But cabal-cache does not error.

Additionally, this doesn't happen on our self hosted runner, only the github runners.

newhoggy:
If downloading the `cabal-cache` does not error on `404` because that happens normally as well, for example using an uninitialised cache.

newhoggy:
You could try using `tmate` to debug Github Actions https://github.com/mxschmitt/action-tmate

hasufell:
> If downloading the `cabal-cache` does not error on `404` because that happens normally as well, for example using an uninitialised cache.

It's on uploading the cache.

newhoggy:
I just tested in Github Actions using `tmate` and it works fine.

I defined `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` and then ran:

```
cabal run -- cabal-cache sync-to-archive '--host-name-override=s3.us-west-004.backblazeb2.com' '--host-port-override=443' '--host-ssl-override=True' --region us-west-1 --archive-uri s3://haskellworks-cabal-cache-4/0000
cabal run -- cabal-cache sync-from-archive '--host-name-override=s3.us-west-004.backblazeb2.com' '--host-port-override=443' '--host-ssl-override=True' --region us-west-1 --archive-uri s3://haskellworks-cabal-cache-4/0000 --store-path moo
```

newhoggy:
Note that even in `sync-to-archive`, the `404` is normal.  What's happening here is that the tool is testing if a cache entry exists first before attempting to upload a cache entry.  If it already exists, the upload is skipped, if not, the upload is performed.

hasufell:
> I just tested in Github Actions using `tmate` and it works fine.
> 
> I defined `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` and then ran:
> 
> ```
> cabal run -- cabal-cache sync-to-archive '--host-name-override=s3.us-west-004.backblazeb2.com' '--host-port-override=443' '--host-ssl-override=True' --region us-west-1 --archive-uri s3://haskellworks-cabal-cache-4/0000
> cabal run -- cabal-cache sync-from-archive '--host-name-override=s3.us-west-004.backblazeb2.com' '--host-port-override=443' '--host-ssl-override=True' --region us-west-1 --archive-uri s3://haskellworks-cabal-cache-4/0000 --store-path moo
> ```

Did you use the prebuilt binaries? If not, how did you build cabal-cache and with what GHC version?

newhoggy:
I used `main` built with `ghc-8.10.7`.

hasufell:
@newhoggy my suspicion is that the pre-build binaries are busted. I can reproduce the 404 errors in tmate with them. When I tried to build from source I got disconnected unfortunately and couldn't connect again even after restarting the action, since it started to throw obscure errors: https://github.com/haskell/ghcup-hs/actions/runs/3584334939/jobs/6031013660#step:3:36

lol... github is a mess

newhoggy:
I have caching working in this Github Actions run from existing binaries: https://github.com/newhoggy/hw-prim/actions/runs/3588997156/jobs/6040968366

newhoggy:
I have added Backblaze to CI to ensure it keeps working: https://github.com/haskell-works/cabal-cache/pull/183

newhoggy:
This Github Action now has support for non-AWS S3 backends: https://github.com/action-works/cabal-cache-s3 if you're interested in using it.

newhoggy:
It downloads it from Github releases at the moment.

newhoggy:
Do you need additional binaries?

newhoggy:
It can also point to a URL to download from an alternate location.

newhoggy:
See: https://github.com/action-works/cabal-cache-s3/blob/main/action.yml

The `cabal-cache-download-uri` config allows you to change to an alternate location.  You would need to use the same convention as Github releases:

See:

* https://github.com/action-works/cabal-cache-s3/blob/main/src/restore.ts#L8-L39
* https://github.com/haskell-works/cabal-cache/releases

newhoggy:
Also, one advantage of using the action (which you can do manually) is that even in the case of a build failure, the action will upload whatever packages did manage to get built so your second run after fixing CI will run faster.

hasufell:
The STM bug is still there

```
cabal-cache: thread blocked indefinitely in an STM transaction
```

https://github.com/haskell/ghcup-hs/pull/694/checks?check_run_id=9863049080

hasufell:
Alright... I tried everything. Debugging in tmate etc. Nothing works reliably. I'm giving up on cabal-cache.

newhoggy:
> The STM bug is still there
> 
> ```
> cabal-cache: thread blocked indefinitely in an STM transaction
> ```
> 
> https://github.com/haskell/ghcup-hs/pull/694/checks?check_run_id=9863049080

I'm not able to see from the above link the error message.  Do you know if it was `sync-from-archive` or `sync-to-archive` that failed in this way?

newhoggy:
You can speed up iteration time by doing `ecabal build --dependencies-only aeson` instead of ecabal build --dependencies-only "$@"` when debugging build issues.  I did this to get a `tmate` session running in reasonable time.

newhoggy:
This seems to be proof this is happening:

```
runner@fv-az554-623:~/work/ghcup-hs/ghcup-hs$ cat dist-newstyle/cache/plan.json | jq . | head
{
  "cabal-version": "3.6.2.0",
  "cabal-lib-version": "3.6.2.0",
  "compiler-id": "ghc-8.10.7",
  "os": "linux",
  "arch": "x86_64",
  "install-plan": [
    {
      "type": "configured",
      "id": "Cabal-3.6.2.0-ff57806f753861324d8a9cc7c35dc44cc111f0e344c881dec2a0db3f3089ed31",
runner@fv-az554-623:~/work/ghcup-hs/ghcup-hs$ ghc --version
The Glorious Glasgow Haskell Compilation System, version 9.4.3
```

newhoggy:
Hmm, that's not it.  In `tmate`, the `ghc` version is different to in the build.

hasufell:
> > The STM bug is still there
> > 
> > ```
> > cabal-cache: thread blocked indefinitely in an STM transaction
> > ```
> > 
> > https://github.com/haskell/ghcup-hs/pull/694/checks?check_run_id=9863049080
> 
> I'm not able to see from the above link the error message.  Do you know if it was `sync-from-archive` or `sync-to-archive` that failed in this way?

```
cabal-cache sync-from-archive --aws-log-level=Debug --host-name-override=s3.us-west-004.backblazeb2.com --host-port-override=443 --host-ssl-override=True --region us-west-1 --archive-uri s3://ghcup-hs
GHC package manager version 8.10.7
Store path: /.cabal/store
Store path hash: c4f822dfda
Archive URI: s3://ghcup-hs
Archive version: v2
Threads: 4
AWS Log level: Just Debug
Creating store directories
Package DB missing. Creating Package DB
Skipping: rts
Skipping: ghc-prim-0.6.1
Skipping: integer-gmp-1.0.3.0
Skipping: base-4.14.3.0
Skipping: array-0.5.4.0
cabal-cache: MissingFileError "//.aws/credentials"
cabal-cache: MissingFileError "//.aws/credentials"
cabal-cache: MissingFileError "//.aws/credentials"
cabal-cache: MissingFileError "//.aws/credentials"
cabal-cache: thread blocked indefinitely in an STM transaction
```

https://cirrus-ci.com/task/4836283567046656?logs=main#L6496

newhoggy:
Try this: https://github.com/haskell/ghcup-hs/pull/703

newhoggy:
I used this to figure out where the cabal store is actually found:

```
cabal help user-config
```

newhoggy:
Hmmm, I don't know why the cabal store is located in `/github/workspace/cabal/store` in the build but when I `tmate` and then run docker, I get `/home/runner/work/ghcup-hs/ghcup-hs/cabal` instead.  I suppose I can just compute it instead.

newhoggy:
By specifying `--store-path` uploads are working now.

Screenshot 2022-12-07 at 1 13 31 am

newhoggy:
This is working in my own private PR https://github.com/newhoggy/ghcup-hs/pull/3

You can pick the changes from there.  You really only need the lines that add the `--store-path` and the lines that compute it.

newhoggy:
Actually, this is the relevant commit: https://github.com/newhoggy/ghcup-hs/pull/3/commits/e7816b3295dd23b11403607fdb755b5b187714ac

newhoggy:
One more thing that can potentially mess up the cache is if anything in `PackageHashConfigInputs` is not reproduced between builds resulting is mismatched package hashes.

https://hackage.haskell.org/package/cabal-install-3.8.1.0/docs/Distribution-Client-PackageHash.html

As long as the inputs are stable between builds the download should work.  I'm mentioning this because I haven't checked yet if downloads are working.

newhoggy:
Looks like there is some hash instability.  For example the build is trying to download:

```
Failed to download: safe-exceptions-0.1.7.3-3a38637f5927c84370651d20ecb77c5817b46445bedb1764caec1bd6d587f838
```

But it uploads:

```
Creating s3://haskellworks-cabal-cache-backblaze/test-cache-3/v2/ghc-8.10.7/safe-exceptions-0.1.7.3-3b3c93f2acca9bea925402b33c9a87bba21e83bc209159aa41606a30bf4eab6b.tar.gz
```

The hash doesn't match.

I don't have insight into why this might be the case but you might.

newhoggy:
Oh it's working now.  I managed to fix it with a `--dry-run`:

https://github.com/newhoggy/ghcup-hs/pull/3/commits/2e713a2efef527221d91fee514c4bbe2a980bfe8

Screenshot 2022-12-07 at 2 07 27 am

newhoggy:
@hasufell I think all the pieces are there for you to be able to get a working `cabal-cache`.  I'm off to sleep now.  Good luck!

hasufell:
> @hasufell I think all the pieces are there for you to be able to get a working `cabal-cache`. I'm off to sleep now. Good luck!

Thank a lot!

Issue #357, opened 2022-04-18
Add support for proper mirrors


hasufell:
In GitLab by @maerwald on Apr 18, 2022, 23:16

Currently, mirrors are custom metadata files, instead of allowing to just change the base url.

The [recent haskell.org outage](https://status.haskell.org/pages/incident/537c07b0cf1fad5830000093/625c69cdfcbdd204d4df5603) shows that it would be useful to have direct mirror support and possibly even allow ghcup to switch automatically.

----

Additionally, ghcup might host its own S3 bucket mirror.

hasufell:
#700 has been merged

hasufell:
This is released in 0.1.19.0

PR #700, opened 2022-12-03
Add support for mirrors wrt #357

hasufell:
@PhotonQuantum ...this might resolve the need for you to modify the metadata file itself. But you will have to replicate the path components. The mirror support allows to

1. replace the host component
2. specify a path prefix

E.g. when we have https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-0.0.7.yaml and the following mirror config:

```yml
mirrors:
  "raw.githubusercontent.com":
    authority:
      host: "mirror.sjtu.edu.cn"
    pathPrefix: "ghcup/yaml"
```

then the resulting url will be https://mirror.sjtu.edu.cn/ghcup/yaml/haskell/ghcup-metadata/master/ghcup-0.0.7.yaml

The current hosts are:
   - github.com (for stack and some older HLS versions)
   - raw.githubusercontent.com (for the yaml metadata)
   - downloads.haskell.org (for everything else)

### DIFF
--- a/.github/scripts/common.sh
+++ b/.github/scripts/common.sh
@@ -151,7 +151,7 @@ install_ghcup() {
 		chmod +x ghcup
 		mv ghcup "$HOME/.local/bin/ghcup"
 	else
-		curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | BOOTSTRAP_HASKELL_NONINTERACTIVE=1 BOOTSTRAP_HASKELL_MINIMAL=1 sh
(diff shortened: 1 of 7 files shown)


PhotonQuantum:
Looks good! btw will the installer script use this config? We'd like to enable our users to use the mirror when installing ghcup itself.

hasufell:
> Looks good! btw will the installer script use this config? We'd like to enable our users to use the mirror when installing ghcup itself.

Right now not, but this could be made an option via an env var, e.g.

```
curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | GHCUP_MIRROR=sjtu sh
```

But first, the directory structure needs to match so we can test this out.

hasufell:
@PhotonQuantum can we give it a test?

PhotonQuantum:
> @PhotonQuantum can we give it a test?

I'm willing to help, but this requires some significant modifications to our codebase, and I'm occupied these days.

I'll have some time available next month and we may start testing this design then. Sorry for the inconvenience!

hasufell:
I'm going to merge this for now. If there are issues, please raise them in a new ticket.

PR #703, opened 2022-12-06
Add store path to cabal-cache sync

newhoggy:


hasufell:
Still doesn't work on windows: https://github.com/haskell/ghcup-hs/actions/runs/3658670103/jobs/6184264267#step:10:20724

I tried both the form `cabal_store_path="/c/cabal"` and `cabal_store_path="C:\\cabal"`. Both does not work. `cabal help user-config` prints `C:\cabal\config`.

newhoggy:
I wonder if it's because something in `C:\cabal\config` causes the store to be somewhere else?

hasufell:
Doesn't look like it:

```
runneradmin@fv-az89-413 MINGW64 /d/a/ghcup-hs/ghcup-hs
# grep store /c/cabal/config
-- store-dir:

runneradmin@fv-az89-413 MINGW64 /d/a/ghcup-hs/ghcup-hs
# ls /c/cabal/store/
ghc-8.10.7
```

hasufell:
The only thing that makes it work is this:

```sh
		rm -rf /c/Users/runneradmin/AppData/Roaming/cabal
		ln -sf /c/cabal /c/Users/runneradmin/AppData/Roaming/cabal
```

hasufell:
But then it doesn't pick up the cache and rebuilds everything.

hasufell:
So the only way to fix it was to set the cabal store dir to the default `$APPDATA/cabal`. Everything else failed.

I also uploaded more binaries: https://downloads.haskell.org/~ghcup/unofficial-bindists/cabal-cache/1.0.5.1/

PR #715, opened 2022-12-18
Improve warning for incompatible HLS/GHC combos

hasufell:
ghcup-screen

@arjunkathuria @david-christiansen

### DIFF
--- a/lib/GHCup/Utils.hs
+++ b/lib/GHCup/Utils.hs
@@ -1270,11 +1270,13 @@ warnAboutHlsCompatibility = do
   case (currentGHC, currentHLS) of
     (Just gv, Just hv) | gv `notElem` supportedGHC -> do
       logWarn $
-        "GHC " <> T.pack (prettyShow gv) <> " is not compatible with " <>
-        "Haskell Language Server " <> T.pack (prettyShow hv) <> "." <> "\n" <>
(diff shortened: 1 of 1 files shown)


david-christiansen:
I think that "Compile from source" sounds intimidating enough that people might not realize how easy it is. What if it said "GHCup can automate the process of compiling HLS from source. See http://... for instructions." instead?

hasufell:
Screenshot_2022-12-19_00-40-01

david-christiansen:
That's really great!

arjunkathuria:
@hasufell 

That's already in the right direction. Perhaps more like.
--


oh, oh! GHC - 9.4.3 doesn't seem to have a ready-made Haskell Language Server - 1.8.0.0 binary out yet.

Haskell LSP / IDE support may not function correctly.

To ensure everything works for you, you can go with any of the options:-
1. Downgrade your GHC version to a version that has ready-made binaries for HLS.
2. Auto compile HLS for your current GHC version using a simple GHCup command. 
     $ ghcup compile hls (relevant versions info)

hasufell:
@arjunkathuria in my experience people often ignore on-screen messages. That's why I try to keep them concise and with clear and simple language. Most users are in a rush when it comes to installation.

arjunkathuria:
@hasufell That is correct, push it.

PR #717, opened 2022-12-19
Integrate with errors.haskell.org

### LINKED ISSUE #434: Integration with errors.haskell.org?
In GitLab by @david-christiansen on Oct 21, 2022, 19:58

[errors.haskell.org](https://errors.haskell.org) presently documents error messages and warnings from GHC, starting with version 9.6.1 once it's released. The design, however, is intended to accommodate any Haskell tooling whatsoever. Would GHCUp be interested in integrating with the site?

What it entails:
 * Each error or warning is assigned a unique code of the form `[GHCUp-12345]`, where "GHCUp" is a namespace (we can talk about what it would be) and "12345" is a number that's unique in that namespace
 * These codes are kept consistent over time, even if the text is rephrased.

As I see it, there'd be the following benefits:
 * Users have an easier time contributing documentation for how to work around various situations that may arise, so more docs may exist
 * Error messages gain a notion of identity that allows them to be rephrased without breaking all docs
 * Interactive tools that invoke GHCup, Stack, Cabal and GHC can link to one consistent, predictable location if an error message occurs, and new users don't have to figure out which tool is the source of the error

The costs would likely be:
 * A one-time cost to put in the codes and update all tests that refer to the string forms of errors
 * An small but real ongoing burden to think about the relationship between messages in older and newer versions to keep the semantics of the codes consistent over time
 * Some user-written documentation will be in a location other than the manual

What do you think? Is this interesting?

hasufell:
TODO:

- [x] include the error codes in the error logs/prints
- [x] pad error codes with leading zeros
- [x] use ansi terminal hyperlinks

Fixes #434

@david-christiansen 

----

Running `ghcup print-app-errors` will yield:

```
# low level errors (1 to 500)
GHCup-1 No compatible platform could be found
GHCup-10 Unable to find a download for the requested version/distro.
GHCup-20 No update available or necessary.
GHCup-40 Unable to figure out the distribution of the host
GHCup-50 The archive format is unknown. We don't know how to extract it.
GHCup-60 The scheme is not supported (such as ftp).
GHCup-70 The scheme is not supported (such as ftp)
GHCup-80 Unable to merge file trees during installation
GHCup-90 Unable to find a tag of a tool
GHCup-100 Unable to find the next version of a tool (the one after the currently set one)
GHCup-110 The tool (such as GHC) is already installed with that version
GHCup-120 The Directory is supposed to be empty, but wasn't
GHCup-130 The required tool is not installed
GHCup-140 Uninstallation failed with leftover files
GHCup-150 An executable was expected to be in PATH, but was not found
GHCup-160 JSON decoding failed
GHCup-170 A file that is supposed to exist does not exist (oops)
GHCup-180 A file already exists that wasn't expected to exist
GHCup-190 The tar directory (e.g. inside an archive) does not exist
GHCup-200 File digest verification failed
GHCup-210 File PGP verification failed
GHCup-220 Unexpected HTTP status error (e.g. during downloads)
GHCup-230 Malformed headers during download
GHCup-240 Not modified HTTP status error (e.g. during downloads).
GHCup-250 The 'Location' header was expected during a 3xx redirect, but not found.
GHCup-260 Too many redirections.
GHCup-270 A patch could not be applied.
GHCup-280 The Tool requirements could not be found.
GHCup-290 The build config is invalid.
GHCup-300 No version is set for tool (but was expected).
GHCup-310 A download was required or requested, but '--offline' was specified.
GHCup-320 Could not find Hadrian build files. Does this GHC version support Hadrian builds?
GHCup-330 A tool is shadowed in PATH.

# high level errors (5000+)
GHCup-5000 A download failed.
GHCup-7000 Installation or setting the tool failed.
GHCup-8000 The build failed.
GHCup-9000 Setting the current version failed.

# true exceptions (500+)
GHCup-500 A parse error occured.
GHCup-510 A list had an unexpected length.
GHCup-520 URL does not have a base filename.

orphans (800+)
GHCup-800 Failed to parse URI.
GHCup-803 Failed to parse URI. Malformed scheme: MissingColon
GHCup-804 Failed to parse URI. Malformed user info.
GHCup-805 Failed to parse URI. Malformed query.
GHCup-806 Failed to parse URI. Malformed fragment.
GHCup-807 Failed to parse URI. Malformed host.
GHCup-808 Failed to parse URI. Malformed port.
GHCup-809 Failed to parse URI. Malformed path.
GHCup-810 Failed to parse URI.
GHCup-820 Archive extraction result.
GHCup-821 Archive result: fatal
GHCup-822 Archive result: failed
GHCup-823 Archive result: warning
GHCup-824 Archive result: retry
GHCup-825 Archive result: Ok
GHCup-826 Archive result: EOF
GHCup-840 A process exited prematurely.
GHCup-841 A process returned a non-zero exit code.
GHCup-842 A process terminated prematurely.
GHCup-843 A process stopped prematurely.
GHCup-844 Could not find PID for this process.
```

The caveat here is that we have error hierarchies...e.g. a "BuildFailed" can encapsulate more errors. And the numbers are not too predictable. So I introduced the `HFErrorProject` class to deal with that, sorta. Not all numbers will be shown in the above list.

### DIFF
--- a/lib/GHCup/Errors.hs
+++ b/lib/GHCup/Errors.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeOperators           #-}
 {-# LANGUAGE FlexibleInstances           #-}
+{-# LANGUAGE RankNTypes           #-}
 
 {-|
 Module      : GHCup.Errors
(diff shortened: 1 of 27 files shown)


hasufell:
This also shows that the error structure may need some improvement. E.g. when you point ghcup to a non-existent yaml metadata, you get:

```
$ ghcup -s file:loll install ghc latest
[ Error ] [GHCup-160] JSON decoding failed with: YAML exception:
[ ...   ] Yaml file not found: loll
[ ...   ] Consider removing loll manually.
[ ...   ]
```

Or when you specify a non-sensical version to install:

```
$ ghcup install ghc tt3315ms  Mon 17:33
[ Error ] [GHCup-10] Unable to find a download for the requested version/distro.
[ Error ] Also check the logs in /home/hasufell/.ghcup/logs
```

hasufell:
The meat of the implementation uses this class:

```hs
class HFErrorProject a where
  eNum :: a -> Int
  eNum _ = eBase (Proxy :: Proxy a)

  eBase :: Proxy a -> Int

  eDesc :: Proxy a -> String
```

And then we can define:

```hs
instance HFErrorProject ArchiveResult where
  eBase _ = 820

  eNum ArchiveFatal  = 821
  eNum ArchiveFailed = 822
  eNum ArchiveWarn   = 823
  eNum ArchiveRetry  = 824
  eNum ArchiveOk     = 825
  eNum ArchiveEOF    = 826

  eDesc _ = "Archive extraction result."
```

`eDesc` and `eBase` are mainly only there to print the above list based on Proxy values.

And we replace `prettyShow` from the Pretty class with:

```hs
prettyHFError :: (Pretty e, HFErrorProject e) => e -> String
prettyHFError e = (("[GHCup-" <> show (eNum e) <> "] ") <>) . prettyShow $ e
```

david-christiansen:
A couple of thoughts about the UI as presented here:

First off, I'm happy to allocate `GHCup` as your namespace specifier :-)

Secondly, message codes on errors.haskell.org are forever. If a message changes its meaning, it needs a new code, and the old one should forever have the same meaning (even if the phrasing changes completely). It looks to me like your categorization system will be fine with this - there's lots of room to expand in each category, and they're not _too_ finely split up. 

We went with uniformly-sampled randomized codes for GHC because it reduced the likelihood of conflicts on long-running branches and wouldn't result in overfitting of a code ontology. I suspect that this just isn't a big issue for GHCup, which probably has few long-running branches and fewer ways that things can go wrong, but I do think it's worth considering.

I think that leading zeros in the codes is more likely to have them sort as expected in the various interfaces where this data may show up. It also makes them take up the same vertical width in a table, which could make it easier to skim for the most significant digits when it's left-aligned.

I also note that `60` and `70` have the same description in your table.

lib/GHCup/Errors.hs
```diff
@@ -34,7 +35,136 @@ import           URI.ByteString
...
+    , let proxy = Proxy :: Proxy HadrianNotFound in format proxy
+    , let proxy = Proxy :: Proxy ToolShadowed in format proxy
+    , ""
+    , "# high level errors (5000+)"
+    , let proxy = Proxy :: Proxy DownloadFailed in format proxy
+    , let proxy = Proxy :: Proxy InstallSetError in format proxy
+    , let proxy = Proxy :: Proxy BuildFailed in format proxy
+    , let proxy = Proxy :: Proxy GHCupSetError in format proxy
+    , ""
+    , "# true exceptions (500+)"
```
david-christiansen, line 91: I don't understand what makes an exception "true" here. Perhaps a different section header would be more clear to more people?

lib/GHCup/Errors.hs
```diff
@@ -34,7 +35,136 @@ import           URI.ByteString
...
+          e     = PStopped "" []
+      in format''' e proxy "A process stopped prematurely."
+    , let proxy = Proxy :: Proxy ProcessError
+          e     = NoSuchPid "" []
+      in format''' e proxy "Could not find PID for this process."
+    ]
+
+
+prettyHFError :: (Pretty e, HFErrorProject e) => e -> String
+prettyHFError e = (("[GHCup-" <> show (eNum e) <> "] ") <>) . prettyShow $ e
```
david-christiansen, line 159: Is this really more clear than
```suggestion
prettyHFError e = ("[GHCup-" <> show (eNum e) <> "] ") <> prettyShow e
```
?

lib/GHCup/Errors.hs
```diff
@@ -34,7 +35,136 @@ import           URI.ByteString
 
 import qualified Data.Map.Strict               as M
 import qualified Data.Text                     as T
-
+import           Data.Data (Proxy(..))
+
+
+
+allHFError :: String
```
david-christiansen, line 42: This seems difficult to maintain over time. But I guess it's no worse than any other documentation.

david-christiansen: I think they used an injective type family to ensure uniqueness of codes in GHC - I wonder whether something like that could happen here, given that you're manually specifying them? Probably not worth it.

hasufell:
> First off, I'm happy to allocate GHCup as your namespace specifier :-)

`GU` may also be possible.

> I suspect that this just isn't a big issue for GHCup, which probably has few long-running branches and fewer ways that things can go wrong, but I do think it's worth considering.

It has had no long running branches so far.

The thing I'm considering is that some error codes are computed at runtime (although they could be computed at compile time too, if we came up with all hierarchy/inner sum-type combinations). So that's why I assign low numbers to low-level errors and high numbers to high-level errors with large gaps to avoid collisions.

Somewhat randomizing this is possible too, though. Not sure I want to invest that much thought though.

lib/GHCup/Errors.hs
david-christiansen: Also, it would be super cool if the code linked to the appropriate URL. Something like `errors.haskell.org/messages/GHCup-YYYY`. If you do this, then I'll improve the 404 page to better invite docs contributions.

This can be done on Unixy systems using the appropriate escape codes: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda  Support is not universal, but it degrades gracefully. `ansi-terminal` supports this and falls back to nothing on Windows: https://hackage.haskell.org/package/ansi-terminal-0.11.4/docs/System-Console-ANSI.html#g:28

lib/GHCup/Errors.hs
```diff
@@ -105,6 +264,10 @@ instance Pretty CopyError where
   pPrint (CopyError reason) =
     text ("Unable to copy a file. Reason was: " ++ reason)
 
+instance HFErrorProject CopyError where
+  eBase _ = 70
+  eDesc _ = "The scheme is not supported (such as ftp)"
```
david-christiansen, line 269: This seems to be an incorrect description

david-christiansen (commented):
OK, I gave a quick read-through of the implementation. Looks quite reasonable, as I'd expect.

There are a few comments on details, and one message that looks like it has the wrong string associated with it.

I wonder whether all those calls to `runLogger $ logError $ T.pack $ prettyHFError e` should be replaced by something like
`logError e = runLogger $ logError $ T.pack $ prettyHFError e`, which would require less find-replacing when details need to change, and produce more readable diffs in the long run.

david-christiansen:
> some error codes are computed at runtime

Can this be stable over time?

lib/GHCup/Errors.hs
hasufell: They have an instance of `Exception`. The others in fact don't. They are just used within `Except` as error type.

hasufell:
> > some error codes are computed at runtime
> 
> Can this be stable over time?

What I mean is this:

```hs
data BuildFailed = forall es . (ToVariantMaybe BuildFailed es, PopVariant BuildFailed es, Pretty (V es), Show (V es)) => BuildFailed FilePath (V es)

data DownloadFailed = forall xs . (ToVariantMaybe DownloadFailed xs, PopVariant DownloadFailed xs, Show (V xs), Pretty (V xs)) => DownloadFailed (V xs)

data HTTPStatusError = HTTPStatusError Int (M.Map (CI ByteString) ByteString)
```

We could have an error: `BuildFailed "build" (DownloadFailed (HTTPStatusError 404 (M.fromList [])))`.

It will compute the error code going inwards and `+` the numbers. Now the idea I actually have is to be able to **refine** the error codes for sum types like `HTTPStatusError`. We start with a base (e.g. `220`) and then may add `221` for a more prominent case (e.g. `HTTPStatusError 500 (M.fromList [])`). It will be stable in the semantic sense, but the user may one day get a different error code. The description will be the same, just more refined. Is that an acceptable approach?

lib/GHCup/Errors.hs
david-christiansen: That sounds like an implementation detail to me, rather than like a semantic fact about the error message itself that would help the user rectify the situation

david-christiansen:
The most important thing is that the documentation for a given code should always make sense. It's fine to make a new code with a more specific meaning that can be documented, but `GHCup-1234` should never change from "bindist 404" to "disk full". It's fine if `GHCUp-1234` starts off being used for all download failures, and then in a later release, `GHCUp-1235` takes over for the download failures that are due to DNS issues. It's also fine to retire a code forever.

hasufell:
Yeah... I think for the nested errors, I will do a pragmatic approach and only document those that are actually observed by users (or me).

hasufell:
Another thought: since all tools part of this effort at the moment are cli tools... would it make sense to unify an interface to print all possible errors via cli? Similar to `<exe> --numeric-version`?

david-christiansen:
What purpose does listing all the errors serve?

hasufell:
> What purpose does listing all the errors serve?

How else would errors.haskell.org start documenting them? By reading the code? New errors could automatically be added on every release if such a cli interface exists.

lib/GHCup/Errors.hs
hasufell: Well... this information isn't really presented to the end user. However, true exceptions in GHCup indeed have special semantics. If one happens, it basically means "oops" and there's something unexpectedly wrong (e.g. an actual bug in GHCup).

david-christiansen:
I had expected that most people would add documentation when they trigger an error that isn't yet documented, and that tools would either include the list in a README or just open issues in the error message index for each of their messages. But this mode makes sense :-)

I realized that links from GHCup to undocumented errors were giving bad UX. So I added a 404 page to the message index that links to the contribution guides when the URL matches the general pattern of messages. That should also help out with this, and make the hyperlinks in GHCup even more useful.

hasufell:
This will be in the next release.

david-christiansen:
This is great! I'm in the process of documenting one of those errors to serve as an example. Do you know what the version number of the next GHCup release will be, so I can put that in the docs?

david-christiansen:
Thanks!

PR #723, opened 2022-12-20
Improve postRm message, fixes #384

### LINKED ISSUE #384: Wrong cabal store location on windows
In GitLab by @Noughtmare on Jun 6, 2022, 01:04

When uninstalling GHC 9.2.2 I noticed this message:

```
[ Info  ] After removing GHC you might also want to clean up your cabal store at: ~/.cabal/store/ghc-<ghcver>
```

On my machine the cabal store is at `C:/cabal/store/ghc-<ghcver>` (I also think it would be nice if the `ghcver` could be filled in automatically in this message).

Can this message be changed on windows?

hasufell:


hasufell:
https://github.com/haskell/ghcup-metadata/pull/52

app/ghcup/GHCup/OptParse/Rm.hs
```diff
@@ -175,8 +178,12 @@ rm rmCommand runAppState runLogger = case rmCommand of
       )
       >>= \case
             VRight vi -> do
-              forM_ (_viPostRemove =<< vi) $ \msg ->
-                runLogger $ logInfo msg
+              -- TODO: deduplicate with brickMain
+              cabalStore <- liftIO (runIdentity . CC.cfgStoreDir <$> CC.readConfig)
```
phadej, line 182: This will fail if config file doesn't exist, which might be the case if user installed just ghc and then removed it?
Not sure how common that would be, just pointing out, maybe ignore IOExceptions for that?

hasufell: Fixed.

PR #732, opened 2022-12-31
Update install.md

brettgilio:
`libffi6` is [only present on oldstable](https://packages.debian.org/buster/libffi6). Update the suggestion to libffi7 (present on stable) or libffi8 (present on testing)

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -38,11 +38,11 @@ Also see [tags and shortcuts](../guide/#tags-and-shortcuts) for more information
 
 ### Linux Debian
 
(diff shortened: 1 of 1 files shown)


hasufell:
This is autogenerated from the metadata https://github.com/haskell/ghcup-metadata/blob/develop/ghcup-0.0.7.yaml#L9

Which already makes that distinction of debian version.

So we need to update the code that generates this markdown instead: https://github.com/haskell/ghcup-metadata/blob/c42bb4a2ffb681a372210ff29b0e00939e281080/ghcup-gen/Generate.hs#L182

Issue #762, opened 2023-01-24
aarch64-linux binary doesn't work on aarch64-linux-deb10


mpickering:
For example see: https://gitlab.haskell.org/ghc/ghcup-ci/-/jobs/1319561

It seems that the aarch64-linux binary is built with a GHC bindist which links against quite a new version of glibc. This means that ghcup fails when using aarch64-deb10 because the version of glibc is too old.

Fixing this might require coordination with GHC releases to provide aarch64-linux binaries linked against an older glibc version but perhaps also an acceptable solution is to tell users that this configuration is not supported by ghcup.

hasufell:
ubuntu focal is used

hasufell:
New release fixes this: https://github.com/haskell/ghcup-hs/releases/tag/v0.1.19.1

PR #768, opened 2023-02-07
Build arm binaries on deb:10 wrt #762

hasufell:


hasufell:
bionic is a no-go. We can try debian10 instead.

PR #773, opened 2023-02-08
Add a known mirror

ksqsf:
USTCLUG (@ustclug) has been operating a [ghcup mirror](https://mirrors.ustc.edu.cn/help/ghcup.html) for some time.

This PR adds this mirror to the list of the known mirrors in the docs.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -163,6 +163,7 @@ ghcup --url-source=https://some-url/ghcup-0.0.6.yaml list
 #### Known mirrors
 
 1. [https://mirror.sjtu.edu.cn/docs/ghcup](https://mirror.sjtu.edu.cn/docs/ghcup)
+2. [https://mirrors.ustc.edu.cn/help/ghcup.html](https://mirrors.ustc.edu.cn/help/ghcup.html)
 
 ### (Pre-)Release channels
 


hasufell:
Cool... I can't open the page though.

How are you operating the mirror? You're serving a metadata file?

Do you know about the new mirror configuration that doesn't require that anymore?

https://github.com/haskell/ghcup-hs/blob/d3a1115b9935cf7f97699b61b431055f121c40ed/data/config.yaml#L96-L121

ksqsf:
Thanks for the new info! I didn't know this new option.

Currently, we serve customized metadata files. The mirroring program reads the official metadata, downloads all artifacts, and rewrite a new metadata file for end users, and the users point `OwnSource` to our versions of metadata. (Not sure why you can't access that page though. It should be public.)

AFAIU, the new method to mirror ghcup is:

- mirrors distribute the *original* metadata files + artifacts
- users configure both `OwnSource` and `mirrors`

Is this understanding right?

hasufell:
They would only need to configure `mirrors`. It will use the official metadata. There's no need to use a custom metadata. However, the mirror would have to replicate the path components now as explained in the config.

Issue #777, opened 2023-02-08
Unable to locate package libffi6


turbotimon:
**Problem**
The [installation page](https://www.haskell.org/ghcup/install/#linux-ubuntu) states for **Debian** and **Ubuntu** that **libffi6** is needed.

However, with the newest stable releases this is most likely **libffi7**, see #219, #93 and [this](https://forums.sourcepython.com/viewtopic.php?f=10&t=2250). libfi6 _can not_ be installed:
```bash
$ sudo apt update && sudo apt install libffi6
Hit:1 http://deb.debian.org/debian bullseye InRelease
Hit:2 http://deb.debian.org/debian-security bullseye-security InRelease
Hit:3 http://deb.debian.org/debian bullseye-updates InRelease
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
9 packages can be upgraded. Run 'apt list --upgradable' to see them.
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package libffi6
```

**Solution**
Therefore, i recommend to change this on the installation page, so it does not confuse people (like it did me). e.g:
> The following distro packages are required: build-essential curl libffi-dev **libffi6 *or* libffi7** libgmp-dev libgmp10 libncurses-dev libncurses5 libtinfo5

**Additional**
Tested with
```
$ cat /etc/*release
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```
and
```
$ cat /etc/*release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.5 LTS"
NAME="Ubuntu"
VERSION="20.04.5 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.5 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
[...]

PR #780, opened 2023-02-11
update Ubuntu required packages: libffi6 -> libffi8ubuntu1

timo-a:
I just installed ghcup on ubuntu 22.04.1 (LTS) and apt-get couldn't find `libffi6`. In addition the installer says:
> Please ensure the following distro packages are installed before continuing (you can exit ghcup and return at any time): build-essential curl libffi-dev libffi8ubuntu1 libgmp-dev libgmp10 libncurses-dev libncurses5 libtinfo5
So I guess these instructions are outdated regarding libffi6.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -42,7 +42,7 @@ The following distro packages are required: `build-essential curl libffi-dev lib
 
 ### Linux Ubuntu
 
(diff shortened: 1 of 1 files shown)


hasufell:
https://github.com/haskell/ghcup-hs/issues/777

This has to be fixed more elaborate. We need to distinguish between ubuntu versions.

PR #781, opened 2023-02-12
Improve usability on 'ghcup config add-release-channel'

### LINKED ISSUE #751: Should `ghcup config add-release-channel` build a set rather than a list?
Should `ghcup config add-release-channel` build a set rather than a list, or at least a nublist (if order matters)?

After hitting `ghcup config add-release-channel ...` twice, I am seeing:
```console
$ ghcup config 
...
url-source:
  AddSource:
  - Right: https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-prereleases-0.0.7.yaml
  - Right: https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-prereleases-0.0.7.yaml
...
```

hasufell:
Fixes #751 (or so I hope).

@andreasabel 

----

Test:

```
ghcup config add-release-channel https://raw.githubusercontent.com/haskell/ghcup-metadata/develop/ghcup-vanilla-0.0.7.yaml
ghcup config add-release-channel https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-prereleases-0.0.7.yaml
ghcup config add-release-channel https://raw.githubusercontent.com/haskell/ghcup-metadata/develop/ghcup-vanilla-0.0.7.yaml
```

Result:

```
[ Error ] [GHCup-00350] Duplicate release channel detected when adding:
[ ...   ]   https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-vanilla-0.0.7.yaml
[ ...   ] Giving up. You can use '--force' to remove and append the duplicate URI (this may change order/semantics).
```

### DIFF
--- a/app/ghcup/GHCup/OptParse/Config.hs
+++ b/app/ghcup/GHCup/OptParse/Config.hs
@@ -159,6 +159,9 @@ updateSettings usl usr =
     --[ Entrypoint ]--
     ------------------
 
+data Duplicate = Duplicate     -- ^ there is a duplicate somewhere in the middle
+               | NoDuplicate   -- ^ there is no duplicate
+               | DuplicateLast -- ^ there's a duplicate, but it's the last element
 
 
 config :: forall m. ( Monad m
@@ -207,16 +210,20 @@ config configCommand settings userConf keybindings runLogger = case configComman
     r <- runE @'[DuplicateReleaseChannel] $ do
       case urlSource settings of
         AddSource xs -> do
-          when (not force && Right uri `elem` xs) $ throwE (DuplicateReleaseChannel uri)
-          lift $ doConfig (defaultUserSettings { uUrlSource = Just $ AddSource (appendUnique xs (Right uri)) })
-          pure ()
+          case checkDuplicate xs (Right uri) of
+            Duplicate
+              | not force -> throwE (DuplicateReleaseChannel uri)
+            DuplicateLast -> pure ()
+            _ -> lift $ doConfig (defaultUserSettings { uUrlSource = Just $ AddSource (appendUnique xs (Right uri)) })
         GHCupURL -> do
           lift $ doConfig (defaultUserSettings { uUrlSource = Just $ AddSource [Right uri] })
           pure ()
         OwnSource xs -> do
-          when (not force && Right uri `elem` xs) $ throwE (DuplicateReleaseChannel uri)
-          lift $ doConfig (defaultUserSettings { uUrlSource = Just $ OwnSource (appendUnique xs (Right uri)) })
-          pure ()
+          case checkDuplicate xs (Right uri) of
+            Duplicate
+              | not force -> throwE (DuplicateReleaseChannel uri)
+            DuplicateLast -> pure ()
+            _ -> lift $ doConfig (defaultUserSettings { uUrlSource = Just $ OwnSource (appendUnique xs (Right uri)) })
         OwnSpec spec -> do
           lift $ doConfig (defaultUserSettings { uUrlSource = Just $ OwnSource [Left spec, Right uri] })
           pure ()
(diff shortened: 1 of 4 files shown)


app/ghcup/GHCup/OptParse/Config.hs
```diff
@@ -203,22 +203,40 @@ config configCommand settings userConf keybindings runLogger = case configComman
...
 
  where
+  -- appends the element to the end of the list, but also removes it from the existing list
+  appendUnique :: Eq a => [a] -> a -> [a]
+  appendUnique xs' e = go xs'
+   where
+    go [] = [e]
+    go (x:xs)
+      | x == e    =     go xs -- skip
+      | otherwise = x : go xs
```
hasufell, line 238: Here is the meat.

app/ghcup/GHCup/OptParse/Config.hs
```diff
@@ -203,22 +203,40 @@ config configCommand settings userConf keybindings runLogger = case configComman
...
-      OwnSource xs -> do
-        doConfig (defaultUserSettings { uUrlSource = Just $ OwnSource (xs <> [Right uri]) })
-        pure ExitSuccess
-      OwnSpec spec -> do
-        doConfig (defaultUserSettings { uUrlSource = Just $ OwnSource ([Left spec, Right uri]) })
+  AddReleaseChannel force uri -> do
+    r <- runE @'[DuplicateReleaseChannel] $ do
+      case urlSource settings of
+        AddSource xs -> do
+          when (not force && Right uri `elem` xs) $ throwE (DuplicateReleaseChannel uri)
```
hasufell, line 210: here we error

app/ghcup/GHCup/OptParse/Config.hs
```diff
@@ -203,22 +203,40 @@ config configCommand settings userConf keybindings runLogger = case configComman
...
+          lift $ doConfig (defaultUserSettings { uUrlSource = Just $ AddSource (appendUnique xs (Right uri)) })
+          pure ()
+        GHCupURL -> do
+          lift $ doConfig (defaultUserSettings { uUrlSource = Just $ AddSource [Right uri] })
+          pure ()
+        OwnSource xs -> do
+          when (not force && Right uri `elem` xs) $ throwE (DuplicateReleaseChannel uri)
```
hasufell, line 217: ditto

cartazio:
conceptually, is the configuration a Map from keys to config values wrapped in the First Monoid https://hackage.haskell.org/package/base-4.17.0.0/docs/Data-Monoid.html#t:First by default, and the --force flag would correspond with the Last monoid? https://hackage.haskell.org/package/base-4.17.0.0/docs/Data-Monoid.html#t:Last

roughly `Map ConfigName (First ConfigValue)`?

hasufell:
@cartazio no idea. Never used those.

`--force` does not have an effect on which keys in the nested Maps are overwritten, it only changes the overwrite order.

Say:

- A: `ghcup-vanilla-0.0.7.yaml`
- B: `ghcup-0.0.7.yaml`

Both have duplicate keys (GHC versions).

If you have configured the order `[ A, B ]`, then B takes precedence. 

If you now say **add A** with current GHCup, you get `[ A, B, A ]` which is the same as `[ B, A ]` just more inefficient. 

With the new patch, ghcup will error out.

With `--force`, you'll get `[ B, A ]` (as opposed to `[ A, B, A ]`).

Does that make sense?

cartazio:
I will admit I’m even more confused about the current semantics then :)

I understand that for non conflicting keys, it should be a union.

What’s the current shadowing override rule on conflict and what’s the new
one?

hasufell:
> What’s the current shadowing override rule on conflict and what’s the new one?

Duplicate keys are overwritten right biased, no matter whether you use `--force` or not. `--force` just changes the element index of one of the elements in the list we union over.

app/ghcup/GHCup/OptParse/Config.hs
andreasabel: So, if I understand correctly, addition sequence `a.b.a` will now error unless `force`d.  
But also `a.a` will error, but this is in effect the same as `a`, because the action of adding a fixed channel at the end is idempotent (if we ignore the extra entry in `xs`).
I am a bit worried about the error in such don't-care situations.  This could lead to regressions for uses of `ghcup` in scripts.  
But I understand that just ignoring duplicate additions with a warning (unless `force`) is also not a backwards-compatible solution, and even worse than a hard error.

Could one stick with a warning just in case the last channel is added again?  In these cases we know that adding it again is redundant.

PR #782, opened 2023-02-17
Doc refactoring and cleanup

chreekat:
It looks like this file was left behind in the mkdocs migration.

A follow up question: should the www/ directory just be removed?


Kleidukos:
Something tells me that this block needs it:

```js
	if (platforms.includes(platform)) {
		platforms.forEach(function (platform_elem) {
			var platform_div = document.getElementById("ghcup-instructions-" + platform_elem);
			platform_div.style.display = "none";
			if (platform == platform_elem) {
				platform_div.style.display = "block";
			}
		});
	}
}
```
(platform_div is null right now)

chreekat:
@Kleidukos extra.js  is still included and seems like it might be the successor to ghcup.js. At least on Linux and Windows, the page still "works" in that it shows the right instructions for both platforms.

Does it also work for you? Maybe platform_div is also vestigial? I'll look to see what it's donig

chreekat:
I do get `Uncaught TypeError: platform_div is null` on Linux, but not on Windows

chreekat:
Ah, I see. extra.js is included twice: once manually in index.md, and once by mkdocs.

The manual inclusion is at the top of the page, before any content has been loaded. That one throws an error (on Firefox on Linux, anyway. I guess not on Chrome on Windows.) The second one presumably does the actual work.

chreekat:
font-awesome is also included twice

chreekat:
Ok, now I've fixed or refactored a few things.

I checked the changes with `mkdoc build` and the only functional differences are:

* New favicon
* Removal of unused ghcup.js
* Removal of duplicate extra.js
* Removal of duplicate font-awesome.css

The refactoring was a more idiomatic implementation of removing the next/prev links.

hasufell:
I'm an idiot when it comes to frontend, so please consider doing radical changes.

chreekat:
Yeah, done for now. :)

hasufell:
merged and deployed

Issue #788, opened 2023-02-20
Can `latest` apply to prereleases too?


ulysses4ever:
Perhaps, not `latest` but something else (`latest-prerelease`?), I don't care about a particular word, I just want a generic name, so that I don't haveto memoize all that 9.6.0.xyzabc business..

hasufell:
We definitely need a succint name for the tag. `latest-prerelease` sounds ok to me.

This will require code changes.

Any motivation for a PR?

ulysses4ever:
I’m afraid I’m overcommitted currently :-(

hasufell:
Released https://github.com/haskell/ghcup-hs/releases/tag/v0.1.19.2

PR #790, opened 2023-02-21
Implement 'latest-prerelease' tag wrt #788

hasufell:
@ulysses4ever 

----

checked this parses with old ghcup


ulysses4ever:
That was fast, thank you!!!

Issue #804, opened 2023-03-07
FreeBSD 13.1 installation fails


avysk:
Apparently, `ghc` depends on `libnursesw.8.so`, which is not available. However, the library can be installed with `compat12x-amd64-12.2.1202000.20210406` package, and it fixes the installation. Until `ghc` is fixed, the installation documentation should be updated.

<img width="658" alt="image" src="https://user-images.githubusercontent.com/130091/223382755-3e39012f-7885-408b-916f-525ebe857c5b.png">

hasufell:
Which GHC?

avysk:
Obviously, the one `ghcup` installs by default: `9.2.5`.
```
❯ ldd ~/.ghcup/ghc/9.2.5/lib/ghc-9.2.5/bin/ghc | grep curses
        libncursesw.so.8 => /usr/local/lib/compat/libncursesw.so.8 (0x8080f6000)
```

hasufell:
Can you try 9.2.7?

avysk:
Sure, a moment.

avysk:
Looks good.
```
❯ ldd ~/.ghcup/ghc/9.2.7/lib/ghc-9.2.7/bin/ghc | grep curses
        libncursesw.so.9 => /lib/libncursesw.so.9 (0x80744b000)
```

hasufell:
9.2.7 will be the default version soon, then it will fix itself

avysk:
The problem is that it probably won't work on FreeBSD 12 which is still a production release according to www.freebsd.org

hasufell:
I only support FreeBSD 13.x and that's already hard as it is, because GHC has no official support anymore. I build the bindists myself.

avysk:
Should the docs mention it is 13 only?

hasufell:
https://www.haskell.org/ghcup/install/#freebsd_1

Feel free to create a PR that updates the description.

avysk:
There was some problem in CI, but surely not caused by my editing of markdown: `.github/scripts/common.sh: line 18: S3_HOST: unbound variable`

PR #805, opened 2023-03-07
Adds some information about FreeBSD installation.

avysk:
Documentation for #804.


docs/install.md
```diff
@@ -16,6 +16,8 @@ For Linux, macOS, FreeBSD or Windows Subsystem 2 for Linux, run this in a termin
 curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
 ```
 
+Notice that only FreeBSD 13.x is supported. If the installation fails, complaining about `libncursesw.8.so`, you will need to install FreeBSD 12 compat package first, for example, `pkg install compat12x-amd64-12.2.1202000.20210406`.
```
hasufell, line 19: The shorthand form is `misc/compat12x`.

PR #809, opened 2023-03-16
Use correct environment variable for `STACK_ROOT` in bootstrap script

Rypac:
The environment variable used to set the root Stack directory was incorrectly defined as `STACK_ROOOT` (typo with an extra 'O'). 

This has been fixed and the correct `STACK_ROOT` variable is now used.

See: https://docs.haskellstack.org/en/stable/environment_variables/#stack_root


hasufell:
Ouch, lol

Rypac:
Yes, I was scratching my head for a while until I noticed the extra 'O' 😂

Not sure what is up with the CI. Any ideas?

hasufell:
Released.

PR #818, opened 2023-04-14
Suggest remove of cabal store only when GHC is uninstalled

### LINKED ISSUE #817: Incorrect message after removing HLS version
```
[ Info  ] Removing files safely from: /home/allbery/.local/share/ghcup/hls/1.9.0.0
[ Info  ] After removing GHC you might also want to clean up your cabal store at: ~/.cabal/store/ghc-1.9.0.0
Success
```
I don't think a message relevant to removing a GHC version belongs there?

lsmor:
Fixes #817 

It is a very simple change. I have tested it and it works.


lsmor:
ping @hasufell

PR #829, opened 2023-05-16
Removed political messaging.

relaybit:
a


hasufell:
I will keep the current colors, unless haskell.org views this as inappropriate (because they host the site).

CCing @tomjaguarpaw @david-christiansen

PR #831, opened 2023-05-26
Rerun hlint --default . > .hlint.yaml for up-to-date counts.

philderbeast:


hasufell:
See https://github.com/haskell/ghcup-hs/pull/831

Issue #847, opened 2023-07-07
Create an "introductory mode" in `ghcup tui`


hasufell:
https://discourse.haskell.org/t/cabal-quickstart-with-hls-pains/6867/8?u=hasufell

Issue #65, opened 2020-09-01
Improve TUI


hasufell:
In GitLab by @maerwald on Sep 2, 2020, 03:33

Currently, when a command is run we set the TUI on hold and execute in the "shell". It would be nicer if everything is in the TUI.

hasufell:
The biggest hurdle here is getting subprocess logging working inside the TUI.

runeksvendsen:
> The biggest hurdle here is getting subprocess logging working inside the TUI.

@hasufell could I get you to note here in this issue which places in the code needs to change to fix this?

Is it only `execLogged` that needs to be changed, or is there more?

hasufell:
> > The biggest hurdle here is getting subprocess logging working inside the TUI.
> 
> @hasufell could I get you to note here in this issue which places in the code needs to change to fix this?
> 
> Is it only `execLogged` that needs to be changed, or is there more?

I don't really know. That has to be investigated.

PR #850, opened 2023-07-11
[Draft] Introductory mode

lsmor:
This is a proof of concept for #847 . It creates a PopUp on instalation, asking if you want to set the installed tool as the default. 

This isn't an introductory mode, but it reflectes how challenging would it be. This PR contains very very ugly code, which would become unmaintanable very quickly. More over creating a complicated introductory mode would require a huge overhaul of the actual code.

- Dependencies on `microlens` are unavoidable is we want decent code
- Modal data type should be include. Something like `data Mode = ShowingDialog | Normal | Introduction | ...`
- Change `appDraw` to depend on the `Mode`. Adding different drawing functions for each `Mode`
- Modify the `evenHandler` the same way (one handler for each mode, and a function to dispatch).

I am willing to do it but I don't know If It worth a huge rewrite.

(Notice, this poc is because you commented "That requires some design and probably a lot more TUI code" in the discord thread... And that stament is very true, I just wanted to give some perspective) 

ghcup_dialog_example

### DIFF
--- a/app/ghcup/BrickMain.hs
+++ b/app/ghcup/BrickMain.hs
@@ -40,7 +40,7 @@ import Brick
 import qualified Brick
 import           Brick.Widgets.Border ( hBorder, borderWithLabel)
 import           Brick.Widgets.Border.Style ( unicode )
-import           Brick.Widgets.Center ( center )
+import           Brick.Widgets.Center ( center, centerLayer )
 import qualified Brick.Widgets.List as L
 import           Brick.Focus (FocusRing)
 import qualified Brick.Focus as F
@@ -241,7 +241,13 @@ sectionListSelectedElement generic_section_list = do
 
 -}
 
-type Name = String
+data Name = AllTools       -- The main list widget
+          | Singular Tool  -- The particular list for each tool
+          | IODialog       -- The pop up when installing a tool
+          | TutorialBox    -- The tutorial widget
+          deriving (Eq, Ord, Show)
+
+data Mode = Navigation | Tutorial deriving (Eq, Show, Ord)
 
 installedSign :: String
 #if IS_WINDOWS
@@ -287,6 +293,7 @@ data BrickState = BrickState
   , _appSettings :: BrickSettings
   , _appState    :: BrickInternalState
   , _appKeys     :: KeyBindings
+  , _mode        :: Mode
   }
   --deriving Show
 
@@ -295,7 +302,7 @@ makeLenses ''BrickState
 keyHandlers :: KeyBindings
             -> [ ( KeyCombination
                  , BrickSettings -> String
-                 , EventM String BrickState ()
+                 , EventM Name BrickState ()
                  )
                ]
 keyHandlers KeyBindings {..} =
@@ -311,6 +318,7 @@ keyHandlers KeyBindings {..} =
     )
   , (bUp, const "Up", appState %= moveCursor 1 Up)
   , (bDown, const "Down", appState %= moveCursor 1 Down)
+  , (Vty.KChar 'x', const "Tutorial", mode .= Tutorial)
   ]
  where
   --hideShowHandler' :: (BrickSettings -> Bool) -> (BrickSettings -> Bool) -> m ()
@@ -337,7 +345,7 @@ showMod :: Vty.Modifier -> String
 showMod = tail . show
 
 
-ui :: AttrMap -> BrickState -> Widget String
+ui :: AttrMap -> BrickState -> Widget Name
 ui dimAttrs BrickState{ _appSettings = as@BrickSettings{}, ..}
   = Brick.padBottom Max
       ( Brick.withBorderStyle unicode
@@ -432,14 +440,43 @@ ui dimAttrs BrickState{ _appSettings = as@BrickSettings{}, ..}
 minHSize :: Int -> Widget n -> Widget n
 minHSize s' = Brick.hLimit s' . Brick.vLimit 1 . (<+> Brick.fill ' ')
 
-app :: AttrMap -> AttrMap -> App BrickState () String
+app :: AttrMap -> AttrMap -> App BrickState () Name
 app attrs dimAttrs =
-  App { appDraw     = \st -> [ui dimAttrs st]
-  , appHandleEvent  = eventHandler
-  , appStartEvent   = return ()
-  , appAttrMap      = const attrs
-  , appChooseCursor = Brick.showFirstCursor
-  }
+  App { appDraw         = drawUI dimAttrs
+      , appHandleEvent  = eventHandler
+      , appStartEvent   = return ()
+      , appAttrMap      = const attrs
+      , appChooseCursor = Brick.showFirstCursor
+      }
+
+
+drawUI :: AttrMap -> BrickState -> [Widget Name]
+drawUI dimAttrs st = 
+  case st ^. mode of
+    Navigation -> [ui dimAttrs st]
+    Tutorial   -> 
+      let tutorial = centerLayer
+                   $ Brick.hLimitPercent 75
+                   $ Brick.vLimitPercent 50
+                   $ Brick.withBorderStyle unicode
+                   $ borderWithLabel (Brick.txt "Tutorial")
+                   $ Brick.vBox 
+                     [ Brick.txt "GHCup is a tool for managing your Haskell tooling."
+                     , center (Brick.txt "--- o ---")
+                     , Brick.txt "This symbol "
+                        <+> Brick.withAttr (Brick.attrName "installed") (Brick.str "✓ ")
+                        <+> Brick.txt " means that the tool is installed but not in used"
+                     , Brick.txt "This symbol "
+                        <+> Brick.withAttr (Brick.attrName "set") (Brick.str "✔✔") 
+                        <+> Brick.txt " means that the tool is installed and in used"
+                     , Brick.txt "This symbol "
+                        <+> Brick.withAttr (Brick.attrName "not-installed") (Brick.str "✗ ")
+                        <+> Brick.txt " means that the tool isn't installed"
+                     , center (Brick.txt "--- o ---")
+                     , Brick.txt "Press Enter to exit the tutorial"
+                     ]
+       in [tutorial, ui dimAttrs st]
+
 
 defaultAttributes :: Bool -> AttrMap
 defaultAttributes no_color = Brick.attrMap
@@ -481,8 +518,14 @@ dimAttributes no_color = Brick.attrMap
     withBackColor | no_color  = \attr _ -> attr `Vty.withStyle` Vty.reverseVideo
                   | otherwise = Vty.withBackColor
 
-eventHandler :: BrickEvent String e -> EventM String BrickState ()
-eventHandler ev = do
+tutorialHandler :: BrickEvent Name e -> EventM Name BrickState ()
+tutorialHandler ev = 
+  case ev of
+    VtyEvent (Vty.EvKey Vty.KEnter _) -> mode .= Navigation
+    _ -> pure ()
+
+navigationHandler :: BrickEvent Name e -> EventM Name BrickState ()
+navigationHandler ev = do
   AppState { keyBindings = kb } <- liftIO $ readIORef settings'
   case ev of
     inner_event@(VtyEvent (Vty.EvKey key _)) ->
@@ -490,6 +533,13 @@ eventHandler ev = do
         Nothing -> void $ Brick.zoom (toLensVL appState) $ handleGenericListEvent inner_event
         Just (_, _, handler) -> handler
     inner_event -> Brick.zoom (toLensVL appState) $ handleGenericListEvent inner_event
+
+eventHandler :: BrickEvent Name e -> EventM Name BrickState ()
+eventHandler ev = do
+  m <- use mode
+  case m of
+    Navigation -> navigationHandler ev
+    Tutorial   -> tutorialHandler ev
 
 -- | Suspend the current UI and run an IO action in terminal. If the
 -- IO action returns a Left value, then it's thrown as userError.
@@ -524,6 +574,7 @@ updateList appD BrickState{..} =
                  , _appData     = appD
                  , _appSettings = _appSettings
                  , _appKeys     = _appKeys
+                 , _mode        = Navigation
                  }
 
 constructList :: BrickData
@@ -538,8 +589,8 @@ constructList appD settings =
 -- | Focus on the tool section and the predicate which matches. If no result matches, focus on index 0
 selectBy :: Tool -> (ListResult -> Bool) -> BrickInternalState -> BrickInternalState
 selectBy tool predicate internal_state =
-  let new_focus = F.focusSetCurrent (show tool) (view sectionListFocusRingL internal_state)
-      tool_lens = sectionL (show tool) 
+  let new_focus = F.focusSetCurrent (Singular tool) (view sectionListFocusRingL internal_state)
+      tool_lens = sectionL (Singular tool) 
    in internal_state
         & sectionListFocusRingL .~ new_focus
         & tool_lens %~ L.listMoveTo 0            -- We move to 0 first
@@ -559,8 +610,8 @@ replaceLR :: (ListResult -> Bool)
           -> BrickInternalState
 replaceLR filterF list_result s =
   let oldElem = s >>= sectionListSelectedElement -- Maybe (Int, e)
-      newVec  =  [(show $ lTool (head g), V.fromList g) | g <- groupBy ((==) `on` lTool ) (filter filterF list_result)]
-      newSectionList = sectionList "GHCupList" newVec 0
+      newVec  =  [(Singular $ lTool (head g), V.fromList g) | g <- groupBy ((==) `on` lTool ) (filter filterF list_result)]
+      newSectionList = sectionList AllTools newVec 0
   in case oldElem of
       Just (_, el) -> selectBy (lTool el) (toolEqual el) newSectionList
       Nothing -> selectLatest newSectionList
@@ -817,6 +868,7 @@ brickMain s = do
                     defaultAppSettings
                     (constructList ad defaultAppSettings Nothing)
                     (keyBindings (s :: AppState))
+                    Navigation
 
           )
         $> ()
--- a/app/ghcup/BrickMain.hs
+++ b/app/ghcup/BrickMain.hs
@@ -192,8 +192,8 @@ handleGenericListEvent (VtyEvent (Vty.EvKey (Vty.KChar '\t') [])) = sectionListF
 handleGenericListEvent (VtyEvent (Vty.EvKey Vty.KBackTab []))     = sectionListFocusRingL %= F.focusPrev
 handleGenericListEvent (MouseDown _ Vty.BScrollDown _ _)          = moveDown
 handleGenericListEvent (MouseDown _ Vty.BScrollUp _ _)            = moveUp
-handleGenericListEvent (VtyEvent (Vty.EvKey Vty.KDown []))     = moveDown
-handleGenericListEvent (VtyEvent (Vty.EvKey Vty.KUp []))       = moveUp
+handleGenericListEvent (VtyEvent (Vty.EvKey Vty.KDown []))        = moveDown
+handleGenericListEvent (VtyEvent (Vty.EvKey Vty.KUp []))          = moveUp
 handleGenericListEvent (VtyEvent ev) = do
     ring <- use sectionListFocusRingL
     case F.focusGetCurrent ring of
@@ -281,10 +281,7 @@ data BrickData = BrickData
 
 makeLenses ''BrickData
 
-data BrickSettings = BrickSettings
-  { _showAllVersions    :: Bool
-  , _showAllTools       :: Bool
-  }
+data BrickSettings = BrickSettings { _showAllVersions :: Bool}
   --deriving Show
 
 makeLenses ''BrickSettings
@@ -455,7 +452,7 @@ drawTutorial =
             , mkTextBox [
                 Brick.hBox [
                   Brick.withAttr recommendedAttr $ Brick.str "recommended"
-                , Brick.txtWrap " tag is based on ..."
+                , Brick.txtWrap " tag is based on community adoption, known bugs, etc... So It makes this version the least experimental"
                 ]
               , Brick.hBox [
                   Brick.withAttr latestAttr $ Brick.str "latest"
@@ -515,11 +512,6 @@ drawKeyInfo KeyBindings {..} =
               Brick.txt "Press "
             , keyToWidget bShowAllVersions
             , Brick.txtWrap " to show older version of each tool"
-            ]
-          , Brick.hBox [
-              Brick.txt "Press "
-            , keyToWidget bShowAllTools
-            , Brick.txtWrap " to ??? "
             ]
           ]
         ]
@@ -618,18 +610,19 @@ keyHandlers KeyBindings {..} =
   , ( bShowAllVersions
     , \BrickSettings {..} ->
        if _showAllVersions then "Don't show all versions" else "Show all versions"
-    , hideShowHandler' (not . _showAllVersions) _showAllTools
+    , hideShowHandler' (not . _showAllVersions)
     )
+  , (bUp, const "Up", Brick.zoom (toLensVL appState) moveUp)
+  , (bDown, const "Down", Brick.zoom (toLensVL appState) moveDown)
   , (KeyCombination (Vty.KChar 'h') [], const "help", mode .= KeyInfo)
   ]
  where
   --hideShowHandler' :: (BrickSettings -> Bool) -> (BrickSettings -> Bool) -> m ()
-  hideShowHandler' f p = do 
+  hideShowHandler' f = do 
     app_settings <- use appSettings
     let 
       vers = f app_settings
-      tools = p app_settings
-      newAppSettings = app_settings & showAllVersions .~ vers & showAllTools .~ tools
+      newAppSettings = app_settings & showAllVersions .~ vers
     ad <- use appData
     current_app_state <- use appState
     appSettings .= newAppSettings
@@ -719,8 +712,7 @@ constructList :: BrickData
               -> Maybe BrickInternalState
               -> BrickInternalState
 constructList appD settings =
-  replaceLR (filterVisible (_showAllVersions settings)
-                           (_showAllTools settings))
+  replaceLR (filterVisible (_showAllVersions settings))
             (_lr appD)
 
 -- | Focus on the tool section and the predicate which matches. If no result matches, focus on index 0
@@ -1015,7 +1007,7 @@ brickMain s = do
 
 
 defaultAppSettings :: BrickSettings
-defaultAppSettings = BrickSettings { _showAllVersions = False, _showAllTools = False }
+defaultAppSettings = BrickSettings { _showAllVersions = False}
 
 
 getGHCupInfo :: IO (Either String GHCupInfo)
--- a/app/ghcup/BrickMain.hs
+++ b/app/ghcup/BrickMain.hs
@@ -7,6 +7,9 @@
 {-# LANGUAGE ViewPatterns      #-}
 {-# OPTIONS_GHC -Wno-unused-record-wildcards #-}
 {-# OPTIONS_GHC -Wno-unused-matches #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE TemplateHaskell   #-}
 
 module BrickMain where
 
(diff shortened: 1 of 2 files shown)


hasufell:
This is pretty cool!

> I am willing to do it but I don't know If It worth a huge rewrite.

The TUI code is not very large, so I'm open to a complete rewrite. But the visual design should mainly stay the same, with the finicky separators... those required me to copy-paste internals and adjust them, see:

* https://github.com/haskell/ghcup-hs/blob/bdbbeb104064f4e3048ea47881bc4749685f8ba4/app/ghcup/BrickMain.hs#L234
* https://github.com/jtdaugherty/brick/blob/2d6f90c62902351e57e9561c66fa32b3742b4660/src/Brick/Widgets/List.hs#L335

The other thing I want to get rid of is this: https://github.com/haskell/ghcup-hs/issues/65
It's not very nice to suspend the entire TUI. It's a shortcut, a hack. We should open a new dialog window, gather the logs from the functions, print them as we go etc.... but that's a lot more work and may also require some ghcup internals to be re-designed, e.g. https://github.com/haskell/ghcup-hs/blob/bdbbeb104064f4e3048ea47881bc4749685f8ba4/lib/GHCup/Prelude/Process/Posix.hs#L74

----

Wrt the actually proposed change... I think we want to avoid popups by default. I could imagine that after ghcup installation, we print a message to start in introductory mode via `ghcup tui --assistant` or so.

lsmor:
> But the visual design should mainly stay the same

Absolutely. I am not planning to change it.

> It's not very nice to suspend the entire TUI. It's a shortcut

Well. I can give it a try, but I suspect this is going to be indeed difficult. Any case since the refactoring should include `Mode` type (following [recommendations](https://github.com/jtdaugherty/brick/issues/476) by brick's author); such a mode could include an `Action` constructor which isn't used, but with a big TODO: "Should be used to stream logs in a dialog box"

> I think we want to avoid popups by default. I could imagine that after ghcup installation, we print a message to start in introductory mode via `ghcup tui --assistant` or so.

I think many people would like to have the `install andThen set` behaviour as the default. Would you consider a `gchup-tui.conf/json/yaml` file with a serialize version of `BrickSettings` (the refactor would include new fields such as `setDefaultOnInstall`), When installing `ghcup` the default value would be `Nothing`, In such a case, a popup shows

```
               would you like to set the installed tool as the default
 -----------------------------------------------------------------------------------------
 | Yes |    | No |    |Yes, and remember this option |    | No, and remember this option |
```

hasufell:
> I think many people would like to have the install andThen set behaviour as the default.

I don't have any evidence for this. It's the first time someone complains about it.

> Would you consider a gchup-tui.conf/json/yaml file with a serialize version of BrickSettings (the refactor would include new fields such as setDefaultOnInstall), When installing ghcup the default value would be Nothing, In such a case, a popup shows

I'm fine with new specific settings, because the average user doesn't need to know about it.

However, I'm not fond of popup questions by default. These should only exist in "introductory mode" or something else.

Another way is to just introduce another key that denotes "install and set".

hasufell:
Any news?

lsmor:
@hasufell . I've been getting used to the code base a little bit more. 

- Is there any reason to not use the `Brick.Widgets.List.GenericList`? 
- Shouldn't `KeyBindings` belong to `BrickSettings` instead of `Settings`?
- The whole `settings' :: IORef AppState` seems a little bit risky. Are you willing to change that? 
- The introductory mode (I'd prefer `tutorial` which is more explicit) can be done adding  constructors to the previously mentioned `Mode` type. It'd be some popups (one or two) explaining the simbols (cross, check mark and double check mark) and some generic information about the tools.

This is doable (but will take time). If we want a decent/maintanable code a full rewrite of the TUI part should be done. The parts we can save from the current code are those which handle the `AppState`: `install'`, `set'`, etc...

As for the ETA. With very few hours a week, I could have it in october or so. Do you want to continue in this draft or to open a new issue?

hasufell:
> Is there any reason to not use the Brick.Widgets.List.GenericList?

I'm not very familiar with brick. I don't do TUI often. The code was reviewed once in its early stages by the brick author, but that was it.

> Shouldn't KeyBindings belong to BrickSettings instead of Settings?

Yeah maybe.

> The whole settings' :: IORef AppState seems a little bit risky. Are you willing to change that?

Yeah, sure. But it's not a multi-threaded environment or anything.

> As for the ETA. With very few hours a week, I could have it in october or so. Do you want to continue in this draft or to open a new issue?

Sure, just keep it here.

lsmor:
A quick update before vacation. I did some deeper research on the code and `brick`'s API. These are the things I'll try to implement in september: 

- [ ] I see there are many functions using `BrickState -> EventM n BrickState ()` type signature. Given that `EventM` implements `MonadState BrickState`.  All such functions can be simplified to just `EventM n BrickState ()`. 

- [ ] By refactoring `BrickInternalState` as a `Data.Map`-like structure, a widget with finicky separators can be drawn easily, _and_ we can re-use `Brick.Widgets.List`; simplifiying code a lot.

- [ ] Using a `Brick.Keybindings.KeyDispatcher` is a good decision. 

- [ ] I've been experimenting with creating a "logging widget" so there is no need to suspend the TUI. The way to achive this is by creating custom event type `type LogLine = Text` and  a `Brick.BChan` holding those events. I haven't try yet but maybe this could be surprisingly easy If the `LoggerConfig.consoleOutter` can be redirected to write into the `BChan`. 

- [ ] Last but not least, the introductory mode is just two or three panels showing important information about the tool.

ghcup.cabal
```diff
@@ -352,6 +352,9 @@ executable ghcup
       , transformers  ^>=0.5
       , unix          ^>=2.7
       , vty           ^>=5.37
+      , microlens     ^>=0.4.13
```
hasufell, line 355: `optics` does not work? Why?

hasufell:
I don't believe in squashing. Organize the commits however you want.

I'll have capacity to review it. If you have links to the brick documentation about changes done, that might help.

lsmor:
I'll be able to drop dependencies on `microlens`. The main problem was function `zoom` (Re-exported by `brick`) which comes from `microlens` and work with its lenses. It turns out that the type is `LensLike ... ->` and `optics` package allows for converting from its own optic representation to `LensLike` optics representation.  Conclusion: there is interoperability with only a few changes in code.

ghcup.cabal
```diff
@@ -352,6 +352,9 @@ executable ghcup
...
+      , microlens-th  ^>=0.4.3
+      , microlens-mtl ^>=0.2.0
```
hasufell, line 357: These deps also need to be added to the `ghcup-optparse` library component.

lsmor:
I have rebase the branch and finalice the important part of the work (that is the tutorial). I also, change the main data structure for a map-like one which integrates better with `brick` and reuses its code.  

tutorial_example

Some changes with respect the scrolling have changed. Also, the tutorial display is a draft. Tomorrow, I'll make comments in the changes, so it is easier to review

app/ghcup/BrickMain.hs
```diff
@@ -59,70 +74,243 @@ import qualified Data.Text.Lazy.Builder        as B
...
 import qualified Data.Vector                   as V
-import System.Environment (getExecutablePath)
+import           System.Environment (getExecutablePath)
 import qualified System.Posix.Process          as SPP
+import           Optics.TH (makeLenses, makeLensesFor)
+import           Optics.State (use)
+import           Optics.State.Operators ( (.=), (%=), (<%=))
+import           Optics.Operators ((.~), (^.), (%~))
+import           Optics.Getter (view)
+import           Optics.Lens (Lens', lens, toLensVL)
```
lsmor, line 84: Since, the app has been _opticfied_, we need all the imports

app/ghcup/BrickMain.hs
```diff
@@ -59,70 +74,243 @@ import qualified Data.Text.Lazy.Builder        as B
...
+
+-}
+
+data GenericSectionList n t e
+    = GenericSectionList
+    { sectionListFocusRing :: FocusRing n                   -- ^ The FocusRing for all sections
+    , sectionListElements :: !(Vector (L.GenericList n t e)) -- ^ A key-value vector
+    , sectionListName :: n                                   -- ^ The section list name
+    }
+
```
lsmor, line 106: you can think of this as a list of list with a FocusRing keeping track on which list you are. It is inspire by brick's [List](https://hackage.haskell.org/package/brick-1.10/docs/Brick-Widgets-List.html) API In other words

- `GenericList` is provided by `Brick`
- `GenericSectionList` is a collection of `GenericList` and the [`FocusRing`](https://hackage.haskell.org/package/brick-1.10/docs/Brick-Focus.html) which allow dispach events to each section.

app/ghcup/BrickMain.hs
```diff
@@ -59,70 +74,243 @@ import qualified Data.Text.Lazy.Builder        as B
...
+            => n                     -- The name of the section list
+            -> [(n, t e)]            -- a list of tuples (section name, collection of elements)
+            -> Int
+            -> GenericSectionList n t e
+sectionList name elements height
+  = GenericSectionList
+  { sectionListFocusRing = F.focusRing [section_name | (section_name, _) <- elements]
+  , sectionListElements  = V.fromList [L.list section_name els height | (section_name, els) <- elements]
+  , sectionListName = name
+  }
```
lsmor, line 126: A (not so) smart constructor for section lists. In theory, the second argument should be a `NonEmpty` list, but I think it doesn't crash if it is empty

app/ghcup/BrickMain.hs
```diff
@@ -59,70 +74,243 @@ import qualified Data.Text.Lazy.Builder        as B
...
+            let current_idx = L.listSelected current_list
+                list_length = current_list & length
+            if current_idx == Just (list_length - 1)
+                then do 
+                    new_focus <- sectionListFocusRingL <%= F.focusNext
+                    case F.focusGetCurrent new_focus of
+                        Nothing -> pure () -- |- Optic.Zoom.zoom doesn't typecheck but Lens.Micro.Mtl.zoom does. It is re-exported by Brick
+                        Just new_l -> Brick.zoom (toLensVL $ sectionL new_l) (Brick.modify L.listMoveToBeginning)
+                else Brick.zoom (toLensVL $ sectionL l) $ Brick.modify L.listMoveDown
+
```
lsmor, line 159: `moveXX` is a little bit more complicated than it should. We need to check if the current index of the _section_ is its last index, if so, then move to the next section. We use [`zoom`](https://hackage.haskell.org/package/brick-1.10/docs/Brick-Types.html#v:zoom) which is part of the `microlenses` package but re-exported by `Brick`. `zoom` allow to dispach events to parts of the state. For example, the last line `Brick.modify L.listMoveDown` we are re-using `Brick`'s build in list so we don't have to define it ourself.

app/ghcup/BrickMain.hs
```diff
@@ -59,70 +74,243 @@ import qualified Data.Text.Lazy.Builder        as B
...
+handleGenericListEvent (MouseDown _ Vty.BScrollUp _ _)            = moveUp
+handleGenericListEvent (VtyEvent (Vty.EvKey Vty.KDown []))     = moveDown
+handleGenericListEvent (VtyEvent (Vty.EvKey Vty.KUp []))       = moveUp
+handleGenericListEvent (VtyEvent ev) = do
+    ring <- use sectionListFocusRingL
+    case F.focusGetCurrent ring of
+        Nothing -> pure ()
+        Just l  -> Brick.zoom (toLensVL $ sectionL l) $ L.handleListEvent ev
+handleGenericListEvent _ = pure ()
+
```
lsmor, line 200: Very easy event handler. I've added `Tab` and `Shift + Tab` to move to next/prev section. Other events are just dispached to `L.handleListEvent`.

app/ghcup/BrickMain.hs
```diff
@@ -59,70 +74,243 @@ import qualified Data.Text.Lazy.Builder        as B
...
+                  in   wacc
+                    <=> render_separator has_focus_list list_name
+                    <=> inner_widget has_focus_list list_name list
+              )
+              Brick.emptyWidget elms
+  where
+    is_focused_section l = section_focus && Just (L.listName l) == F.focusGetCurrent focus
+    inner_widget has_focus k l = 
+      let w = render_border has_focus k (Brick.vLimit (length l) $ L.renderList render_elem has_focus l)
+       in if has_focus then Brick.visible w else w
```
lsmor, line 225: rendering uses [`Brick.Widget.List.renderList`](https://hackage.haskell.org/package/brick-1.10/docs/Brick-Widgets-List.html#v:renderList) instead of define it ourself. The separators between sections are provided in the first argument. Previously, this was done by hand. Notice that the `viewport` follows the section, not the element in the list. This is a behavioural change wrt current `master`

Btw, the second argument is currently empty, but allow to define borders (not separators) to each section.

app/ghcup/BrickMain.hs
```diff
@@ -59,70 +74,243 @@ import qualified Data.Text.Lazy.Builder        as B
...
+
+{- GHCUp State
+
+-}
+
+data Name = AllTools       -- The main list widget
+          | Singular Tool  -- The particular list for each tool
+          | IODialog       -- The pop up when installing a tool
+          | TutorialBox    -- The tutorial widget
+          deriving (Eq, Ord, Show)
```
lsmor, line 243: It is a good practise to have an explicit `Name` type enumerating all your widgets... It isn't always necessary but still...

app/ghcup/BrickMain.hs
```diff
@@ -59,70 +74,243 @@ import qualified Data.Text.Lazy.Builder        as B
...
 
+data Mode = Navigation | Tutorial deriving (Eq, Show, Ord)
+
```
lsmor, line 245: The `Mode` data type is the [way to go](https://github.com/jtdaugherty/brick/issues/476#issuecomment-1629151920) for `Brick` applications. It isn't part of the fundamentals of brick, but any complex example use it as a sort of "state machine" type

app/ghcup/BrickMain.hs
```diff
@@ -131,167 +319,159 @@ showKey Vty.KUp = "↑"
...
+      <+> minHSize 15 (Brick.str "Version")
+      <+> Brick.padLeft (Pad 1) (minHSize 25 $ Brick.str "Tags")
+      <+> Brick.padLeft (Pad 5) (Brick.str "Notes")
+  renderList' bis = 
+    let allElements = V.concatMap L.listElements $ sectionListElements bis
+        minTagSize = V.maximum $ V.map (length . intercalate "," . fmap tagToString . lTag) allElements
+        minVerSize = V.maximum $ V.map (\ListResult{..} -> T.length $ tVerToText (GHCTargetVersion lCross lVer)) allElements
+        render_separator _ _ = hBorder
+        render_border _ _ = id
+    in Brick.withDefAttr L.listAttr $ renderSectionList render_separator render_border (renderItem minTagSize minVerSize) True bis
```
lsmor, line 350: Here we use `renderSectionList` defined in the top part of the file

app/ghcup/BrickMain.hs
```diff
@@ -131,167 +319,159 @@ showKey Vty.KUp = "↑"
...
-                then withDefAttr listSelectedFocusedAttr
-                else withDefAttr listSelectedAttr
-              makeVisible' = if isSelected then visible . selItemAttr else id
-          in  addSeparator $ makeVisible' elemWidget
-
-      in render
-        $ viewport "GHCup" Vertical
-        $ vBox
-        $ V.toList drawnElements
-
```
lsmor, line 261: We can remove this function completely as `renderSectionList` do its job

app/ghcup/BrickMain.hs
```diff
@@ -131,167 +319,159 @@ showKey Vty.KUp = "↑"
...
+                        <+> Brick.withAttr (Brick.attrName "set") (Brick.str "✔✔") 
+                        <+> Brick.txt " means that the tool is installed and in used"
+                     , Brick.txt "This symbol "
+                        <+> Brick.withAttr (Brick.attrName "not-installed") (Brick.str "✗ ")
+                        <+> Brick.txt " means that the tool isn't installed"
+                     , center (Brick.txt "--- o ---")
+                     , Brick.txt "Press Enter to exit the tutorial"
+                     ]
+       in [tutorial, ui dimAttrs st]
+
```
lsmor, line 453: `drawUI` now splits, depending on which mode we are. This is the reason `Mode` type is so usefull. We can define handlers, drawings, etc... for each mode with barely extra complexity

app/ghcup/BrickMain.hs
```diff
@@ -303,95 +483,96 @@ defaultAttributes no_color = attrMap
...
+  [ (Brick.attrName "active"    , Vty.defAttr `withBackColor` Vty.blue) -- has no effect ??
+  , (Brick.attrName "no-bindist", Vty.defAttr `Vty.withStyle` Vty.dim)
   ]
   where
     withBackColor | no_color  = \attr _ -> attr `Vty.withStyle` Vty.reverseVideo
                   | otherwise = Vty.withBackColor
 
-eventHandler :: BrickState -> BrickEvent String e -> EventM String BrickState ()
-eventHandler st@BrickState{..} ev = do
+tutorialHandler :: BrickEvent Name e -> EventM Name BrickState ()
```
lsmor, line 495: We define a handler for each mode

app/ghcup/BrickMain.hs
```diff
@@ -303,95 +483,96 @@ defaultAttributes no_color = attrMap
...
+        Nothing -> void $ Brick.zoom (toLensVL appState) $ handleGenericListEvent inner_event
+        Just (_, _, handler) -> handler
+    inner_event -> Brick.zoom (toLensVL appState) $ handleGenericListEvent inner_event
 
+eventHandler :: BrickEvent Name e -> EventM Name BrickState ()
+eventHandler ev = do
+  m <- use mode
+  case m of
+    Navigation -> navigationHandler ev
+    Tutorial   -> tutorialHandler ev
```
lsmor, line 517: `eventHandler` dispaches the event depending on the `Mode`

app/ghcup/BrickMain.hs
```diff
@@ -303,95 +483,96 @@ defaultAttributes no_color = attrMap
...
+        settings <- readIORef settings'
+        flip runReaderT settings $ action (curr_ix, e) >>= \case
+          Left  err -> liftIO $ putStrLn ("Error: " <> err)
+          Right _   -> liftIO $ putStrLn "Success"
+        getAppData Nothing >>= \case
+          Right data' -> do
+            putStrLn "Press enter to continue"
+            _ <- getLine
+            pure (updateList data' as)
+          Left err -> throwIO $ userError err
```
lsmor, line 538: I've tried to make this part of the TUI, instead of `SuspendAndResume`. You can find a decent attempt [here](https://github.com/lsmor/ghcup-hs/tree/log-event). It isn't difficult, as it is only to extent the `Mode` type and to add a custom event channel, but there are some difficulties

lsmor (commented):
This commits have some behaviour changes:

- The viewport isn't as smooth as before. It keeps the whole tool section on screen. Not the end of the world 
- Tab and Shift + Tab can be use to move from one tool to another. This can be changed with two lines of code, but I find it useful
- pressing x open the tutorial screen. 
- pressing enter when the tutorial screen is open, closes it
- when in tutorial, no other key works other than Enter.

app/ghcup/BrickMain.hs
hasufell: Yes, the logging and receiving the process output is the main issue. It's tricky with the current code.

lsmor: well the logging isn't a problem. It is relativaly easy to deviate the logging part to a Brick channel so It it can be used as a widget. The main problem is the output of external processes, for example the call to `curl` prints directly on top of the TUI... I don't think this feature will be implemented in this PR

hasufell: > The main problem is the output of external processes, for example the call to curl prints directly on top of the TUI... I don't think this feature will be implemented in this PR

That is what I meant with "logging". On the linux side we use low-level handling of stdout/stderr file descriptors... they could be exposed and used as callback style thing:

https://github.com/haskell/ghcup-hs/blob/9d8d6e329356bd9a9a27dd4628520284d7d97b2e/lib/GHCup/Prelude/Process/Posix.hs#L110-L115

On windows side we use the process package fake pipes:

https://github.com/haskell/ghcup-hs/blob/9d8d6e329356bd9a9a27dd4628520284d7d97b2e/lib/GHCup/Prelude/Process/Windows.hs#L161-L184

So a solution would probably have to write a wrapper around both that exposes the necessary parts and a callback that listens to messages on the pipe/Fd and reacts to it.

hasufell:
I tried to build and run this and moving through the list of elements in the list has some odd behavior. I'm not sure how to explain.

E.g.:

* when I select the first GHC version, the screen starts scrolling down, although that element is in the very middle of the visible list
* when I toggle "show all versions", the currently selected element completely goes out of the visible area
* when I'm at the bottom of the list... I can still press "down" and after 9 key presses, it will jump to the top most element

I also noticed the horizontal separator is missing under the "Tool Version ..." header

lsmor:
I can reproduce with a big enough font (or an small enough terminal). This is happening because the scroll follows the whole section instead of the element in the section. I should change that. 



> * when I'm at the bottom of the list... I can still press "down" and after 9 key presses, it will jump to the top most element

Well, this is happening for the same thing. Essentially you have 9 items out of view... 



> I also noticed the horizontal separator is missing under the "Tool Version ..." header

upps.. already fixed in my local machine. 

Do you have any preference about "tutorial" part? the current text is rather poor (in essence a Mock), but any change is easy to implement.

hasufell:
What's the status of this PR?

lsmor:
> What's the status of this PR?

This PR implements the tutorial / introductory mode. A major effort has gone into refactoring the code so It follows a more `brick`ish approach (with `Mode` type and using built-in Brick data structures). Currently, Everything should work as before but when pressing 'x' a Tutorial widget pops ups. The branch has been rebased with current master.

hasufell:
Another possible topic for introductory mode: https://github.com/haskell/ghcup-metadata/issues/143

lsmor:
Finally, navigation behaves like before. The only difference is that it loops in both ends (pressing down in the last elemente will go to the first, same in the other direction), which I don't think is an issue but I can change it if requested.

ghcup_example

Notice I've rebased master (again) and squashed a few dirty commits.  I see in the CI that windows job isn't working but I can't test on windows.

Comming back to the purpose of this PR: 

- [X] Create `Mode` type so swap between tutorial and navigation
- [X] Create `Name` type to better distinghish between widgets
- [X] Create tutorial Widget
- [ ]  Include windows symbols in the tutorial widget
- [ ]  Add configurable keybinding for tutorial (currently, hardcoded as 'x')
- [ ]  Include information about `recommended`, `hls-powered`, etc... as discussed in haskell/ghcup-metadata#143

hasufell:
Excellent. I'll be in hospital tomorrow and then on sick leave, so I'm not sure when I get around for a proper review

lsmor:
> Excellent. I'll be in hospital tomorrow and then on sick leave, so I'm not sure when I get around for a proper review

No rush!, I'll continue with the three unchecked items. Take as much time as you need

hasufell:
> Add configurable keybinding for tutorial (currently, hardcoded as 'x')

How about we use `?` or the `h` (for help) keys... then we show the keybindings more verbosely and have a dialog allowing you to choose between `Close` and `Show tutorial`?

That way we can also design a proper help page. And I think it's natural to find the tutorial under a help page?

lsmor:
> How about we use `?` or the `h` (for help) keys... then we show the keybindings more verbosely and have a dialog allowing you to choose between `Close` and `Show tutorial`?

I have added a new info layer with a verbose description of the keybindings. I don't know what `t` is actually for. Go back to navigation pressing `q` and go to next layer (tutorial) pressing `Enter`. Both are hardcoded at the moment. I also have improved the tutorial section, I have some doubts about when a tool is recommended.  (You exit with harcoded `q`)

With respect to Keybindings. 

- Should we add configurable keys for the new widgets (open and close them)? 
- should we use bQuit instead of harcoded-q to quit both KeyInfo and Tutorial?
- using `?` might be problematic. In spanish keyboard (for example) it is Shift + apostrophe, I have tested and seems to work well, but maybe other internationalization settings wouldn't work... don't know

ScreenShots for each new widgets:

### Key Info
image

### Tutorial
image

hasufell:
I guess this is mostly ready to merge then. If you mark the PR as ready, I'll try to go over the code some time.

lsmor:
Notice that there are two items with missing documentation.

- **recommended tag**. Up to ghcup web page: _is at the discretion of the GHCup maintainers and based on community adoption (hackage libraries, tools like HLS, stackage support, etc.) and known bugs_. Should I use that directly or there is something more to be consider? I think this tag could be a source of confusion, so I'd rather prefer comfirm the companion text
- **usage of t key**: it does nothing in my computer, and as far as I know, the only tools ghcup is able to install are the one in display.

Best

hasufell:
> Notice that there are two items with missing documentation and I don't now exactly what to put there
> 
> * **recommended tag**. Up to ghcup web page: _is at the discretion of the GHCup maintainers and based on community adoption (hackage libraries, tools like HLS, stackage support, etc.) and known bugs_. Should I use that directly or there is something more to be consider? I think this tag could be a source of confusion, so I'd rather prefer comfirm the companion text

We shouldn't go into too  much detail here but just say this is the most reasonable choice for people starting out with haskell or companies that are not keen to experiment.

> * **usage of t key**: it does nothing in my computer, and as far as I know, the only tools ghcup is able to install are the one in display.
> 

That key is removed in master already (unfortunately, you'll have to rebase due to that... if it's too much trouble, I can revert it temporarily).

lsmor:
> That key is removed in master already (unfortunately, you'll have to rebase due to that... if it's too much trouble, I can revert it temporarily).

Don't worry, I'll rebase it and include the missing recommended doc

hasufell:
I noticed the TUI still doesn't behave like it used to wrt the viewport.

Old:

Peek 2024-01-22 17-08

New:

Peek 2024-01-22 17-10

----

As you can see. Once you're at an offset higher than the render limit of the terminal... your "cursor" will stay at the bottom of the screen. That's not the behavior of e.g. the old GHCup or other terminal TUIs like tig.

I tried to fix it in the code, but don't see an evident way how to do it.

hasufell:
We basically need a way to store the current applied viewport translation and  then calculate whether the new offset would already fall into that visible viewport when applying the previous translation.

I forgot how the previous implementation did it.

hasufell:
I also don't see a way to query Brick "which list elements are currently visible?". If that was possible, we can infer the current translation that is applied.

hasufell:
The old implementation simply relied on [visible](https://hackage.haskell.org/package/brick-2.3.1/docs/Brick-Widgets-Core.html#v:visible).

I tried the same with the current code (we have to apply visible to both the section list and the selected list element):

```diff
@@ -213,30 +216,32 @@ renderSectionList render_elem section_focus (GenericSectionList focus elms sl_na
         let -- A section is focused if the whole thing is focused, and the inner list has focus
             section_is_focused l = section_focus && (Just (L.listName l) == F.focusGetCurrent focus)
             -- We need to limit the widget size when the length of the list is higher than the size of the terminal
-            limit = min (Brick.windowHeight c) (Brick.availHeight c)
             s_idx = fromMaybe 0 $ V.findIndex section_is_focused elms
-            render_inner_list has_focus l = Brick.vLimit (length l) $ L.renderList (\b -> render_elem (b && has_focus)) has_focus l
-            (widget, off) =
-                V.ifoldl' (\wacc i list ->
+            render_inner_list has_focus l = Brick.vLimit (length l) $ L.renderList (\b -> let v = b && has_focus in makeVisible' v . render_elem $
+            makeVisible' b = if b then Brick.visible else id
+            widget =
+                V.ifoldl' (\(!acc_widget) i list ->
                                 let has_focus_list = section_is_focused list
-                                    (!acc_widget, !acc_off) = wacc
-                                    new_widget = if i == 0 then render_inner_list has_focus_list list else hBorder <=> render_inner_list has_focu$
-                                    new_off
-                                        | i < s_idx  = 1 + L.listItemHeight list * length list
+                                    new_widget = if i == 0
+                                                 then makeVisible' has_focus_list $ render_inner_list has_focus_list list
+                                                 else hBorder <=> (makeVisible' has_focus_list $ render_inner_list has_focus_list list)
+                                    new_off
+                                        | i < s_idx  = 1 + L.listItemHeight list * length list
                                         | i == s_idx = 1 + L.listItemHeight list * fromMaybe 0 (L.listSelected list)
                                         | otherwise  = 0
-                                in  (acc_widget <=> new_widget, acc_off + new_off)
+                                in acc_widget <=> new_widget
                                 )
-                (Brick.emptyWidget, 0)
+                Brick.emptyWidget
                 elms
-        Brick.render $  Brick.viewport sl_name Brick.Vertical $ Brick.translateBy (Brick.Location (0, min 0 (limit-off))) widget
+        Brick.render $ Brick.viewport sl_name Brick.Vertical widget
```

This somewhat restores the old behavior, but that has two issues:

* in order to make a whole section visible, you'll get a small sudden "bump" when switching elements between sections.
* if the section is bigger than the whole viewport, it doesn't work at all

lsmor:
Let me check this. I remmember, reliying on visible did not work, you have no manually calculate the viewport tranlation

hasufell:
We probably need https://hackage.haskell.org/package/brick-2.3.1/docs/Brick-Widgets-Core.html#v:visibleRegion

Just `visible` will try to make the upper left corner of the widget visible, which doesn't work if it's a section and we are past the half of the list items.

I think computing the relative location of the selected list item in the section should be possible. That may fix both issues.

hasufell:
```hs
-- This re-uses Brick.Widget.List.renderList
renderSectionList :: (Traversable t, Ord n, Show n, Eq n, L.Splittable t, Semigroup (t e))
                  => (Bool -> e -> Widget n)             -- ^ Rendering function of the list element, True for the selected element
                  -> Bool                                -- ^ Whether the section list has focus
                  -> GenericSectionList n t e            -- ^ The section list to render
                  -> Widget n
renderSectionList render_elem section_focus ge@(GenericSectionList focus elms sl_name) =
    Brick.Widget Brick.Greedy Brick.Greedy $ do
        c <- Brick.getContext
        let -- A section is focused if the whole thing is focused, and the inner list has focus
            section_is_focused l = section_focus && (Just (L.listName l) == F.focusGetCurrent focus)
            render_inner_list has_focus l = Brick.vLimit (length l) $ L.renderList (\b -> let v = b && has_focus in render_elem v) has_focus l
            widget =
                V.ifoldl' (\(!acc_widget) i list ->
                                let has_focus_list = section_is_focused list
                                    (c', r') = case sectionListSelectedElement ge of
                                                 Nothing -> (2, 0)
                                                 Just (selElIx, _) -> (2, selElIx)
                                    new_widget = if i == 0
                                                 then makeVisible' has_focus_list (c', r') $ render_inner_list has_focus_list list
                                                 else hBorder <=> (makeVisible' has_focus_list (c', r') $ render_inner_list has_focus_list list)
                                in acc_widget <=> new_widget
                                )
                Brick.emptyWidget
                elms
        Brick.render $ Brick.viewport sl_name Brick.Vertical widget
 where
  makeVisible' b (c, r) p
    | b = Brick.Widget (hSize p) (vSize p) $ do
            result <- render p
            let imageSize = ( result ^. lensVL Brick.imageL % to Vty.imageWidth
                            , result ^. lensVL Brick.imageL % to Vty.imageHeight - r
                            )
            return $ if imageSize^._1 > 0 && imageSize^._2 > 0
                     then result & (lensVL Brick.visibilityRequestsL) %~ (Brick.VR (Brick.Location (c, r)) imageSize :)
                     else result
    | otherwise = p
```

This comes very close. But there's still some rough edges.

hasufell:
Ah... got it.... the above is 90% there, just had to change

```
                     then result & (lensVL Brick.visibilityRequestsL) %~ (Brick.VR (Brick.Location (c, r)) imageSize :)
```

to

```
                     then result & (lensVL Brick.visibilityRequestsL) %~ (Brick.VR (Brick.Location (c, r)) (1, 1) :)
```

So it always focuses on a tiny dimension. That restores previous behavior.

Issue #180, opened 2021-06-30
Confusing `ghcup set` error


hasufell:
In GitLab by @callanmcgill on Jul 1, 2021, 24:44

When I run `ghcup set` it doesn't seem to be able to find the versions of GHC installed. For example I have the following:
```
> ghcup install ghc-8.10.4
[ Warn  ] This is an old-style command for installing GHC. Use 'ghcup install ghc' instead.
[ Warn  ] GHC ver 8.10.4 already installed; if you really want to reinstall it, you may want to run 'ghcup rm ghc 8.10.4' first
> ghcup set ghc-8.10.4
[ Warn  ] This is an old-style command for setting GHC. Use 'ghcup set ghc' instead.
[ Error ] The version "ghc-8.10.4" of the tool "GHC" is not installed.
```
Any ideas what might be causing this? I also find the warnings very confusing since the commands shown are also on the gitlab readme.

July541:
I met this when I first time used `ghc set`.

For the command like `ghc set ghc-9.2.8`, should print
```
Did you mean `ghc set ghc 9.2.8`?

View all details with `ghc set --help`
```

And for the command with definitely invalid params like `ghc set abcd`, should print
```
Option `option` not available.

View all details with `ghc set --help`
```

Both of the two should die after printing.

/cc @hasufell make sense?

hasufell:
Sorry, I can't follow. We use optparse-applicative for CLI parsing and as such will get error messages that resemble the parser behavior. 

The old parsing is kept for backwards compat. If you have an idea to make the error messages better, please provide a PR.

July541:
Sorry I lost some context, but if we don't do modifications for constructors, there wont be a compatible issue right?

hasufell:
The issue here is really that `ghc-8.4.4` is a valid **version** according to the version parser we use: https://hackage.haskell.org/package/versions-6.0.1/docs/Data-Versions.html#t:Versioning

See:

```
ghci> versioning "ghc-9.2.8"
Right
    ( General
        ( Version
            { _vEpoch = Nothing
            , _vChunks = Chunks
                ( Alphanum "ghc" :| [] )
            , _vRel = Just
                ( Release
                    ( Numeric 9 :|
                        [ Numeric 2
                        , Numeric 8
                        ]
                    )
                )
            , _vMeta = Nothing
            }
        )
    )
```

[...]

PR #861, opened 2023-07-19
GHCup set command error improvement

July541:
This is a demo for #180.

The pr has no effect for command like `ghcup set ghc xxx`, `ghcup set cabal xxx`.

For old-style commands likes `ghcup set xxx`, it has two results. One can set ghc properly, like `ghcup set latest` and `ghcup set recommended`, and others are not, likes `ghcup set ghc-9.2`, it outputs looks like the following even if I've ghc-9.2.8 installed.

```shell
 ~ ghcup set ghc-9.2
[ Warn  ] This is an old-style command for setting GHC. Use 'ghcup set ghc' instead.
[ Warn  ] Assuming you meant version 9.2.8
[ Error ] [GHCup-00130] The version ghc-9.2.8 of the tool ghc is not installed.
```

So my solution is:

1. For scenario one, gives a warning and continue the `set` operation since they might be valid and otherwise gives an error and stop the follow-up steps.

2. For compatibility considerations, the `optparse` section is extracted to test every potential command that has the same effect during this pr.

Now it looks like:

BEFORE:

```shell
~ ghcup set ghc-9.2
[ Warn  ] This is an old-style command for setting GHC. Use 'ghcup set ghc' instead.
[ Warn  ] Assuming you meant version 9.2.8
[ Error ] [GHCup-00130] The version ghc-9.2.8 of the tool ghc is not installed.
```

AFTER:

```shell
~ cabal run ghcup -- set ghc-9.2
[ Error ] Did you mean `ghcup set ghc 9.2`?
[ ...   ]
[ ...   ]   View all candidates with `ghcup set --help`
[ ...   ]
```

BEFORE:

```shell
~ ghcup set
[ Warn  ] This is an old-style command for setting GHC. Use 'ghcup set ghc' instead.
[ Info  ] GHC 9.2.8 successfully set as default version
```

AFTER:

```shell
~ cabal run ghcup -- set 
[ Warn  ] This is an old-style command for setting GHC. Use `ghcup set ghc recommended` instead.
[ Info  ] GHC 9.2.8 successfully set as default version
```

BEFORE:

```shell
~ ghcup set latest
[ Warn  ] This is an old-style command for setting GHC. Use 'ghcup set ghc' instead.
[ Info  ] GHC 9.6.2 successfully set as default version
```

AFTER:

```shell
~ cabal run ghcup -- set latest
[ Warn  ] This is an old-style command for setting GHC. Use `ghcup set ghc latest` instead.
[ Info  ] GHC 9.6.2 successfully set as default version
```

BEFORE:

```shell
~ ghcup set ghc-javascript-unknown-ghcjs-9.6
[ Warn  ] This is an old-style command for setting GHC. Use 'ghcup set ghc' instead.
[ Warn  ] Assuming you meant version 9.6.2
[ Error ] [GHCup-00130] The version ghc-javascript-unknown-ghcjs-9.6.2 of the tool ghc is not installed.
```

AFTER:

```shell
~ cabal run ghcup -- set ghc-javascript-unknown-ghcjs-9.6
[ Error ] Did you mean `ghcup set ghc javascript-unknown-ghcjs-9.6`?
[ ...   ]
[ ...   ]   View all candidates with `ghcup set --help`
[ ...   ]
```

----

@hasufell  I'm not sure if it is worth it just for a more readable prompt. It's more appreciable if you could take a look at this before I continue improvement so I can cut losses early maybe:)

### DIFF
--- a/app/ghcup/GHCup/OptParse/Common.hs
+++ b/app/ghcup/GHCup/OptParse/Common.hs
@@ -82,7 +82,7 @@ data SetToolVersion = SetGHCVersion GHCTargetVersion
                     | SetToolTag Tag
                     | SetToolDay Day
                     | SetRecommended
-                    | SetNext deriving Show
+                    | SetNext deriving (Eq, Show)
 
 prettyToolVer :: ToolVersion -> String
 prettyToolVer (GHCVersion v')  = T.unpack $ tVerToText v'
--- a/app/ghcup/GHCup/OptParse/Common.hs
+++ b/app/ghcup/GHCup/OptParse/Common.hs
@@ -82,7 +82,7 @@ data SetToolVersion = SetGHCVersion GHCTargetVersion
                     | SetToolTag Tag
                     | SetToolDay Day
                     | SetRecommended
-                    | SetNext
+                    | SetNext deriving Show
 
 prettyToolVer :: ToolVersion -> String
 prettyToolVer (GHCVersion v')  = T.unpack $ tVerToText v'
--- a/app/ghcup/GHCup/OptParse/Set.hs
+++ b/app/ghcup/GHCup/OptParse/Set.hs
@@ -249,6 +249,61 @@ runSetStack runAppState =
     --[ Entrypoints ]--
     -------------------
 
+-- | Given a old-style @SetToolVersion@ and give a recommendation with
+-- maximum probability.
+-- To maintain compatable, returns @m True@ to indicate we can continue
+-- do a @set@ operataion, otherwise @m False@ to cut early, these should
+-- be failed even if we continue do @set@.
+guessTargetForSetCmd :: forall m.
+    ( Monad m
+    , MonadMask m
+    , MonadUnliftIO m
+    , MonadFail m
+    )
+  => (ReaderT LeanAppState m () -> m ())
+  -> SetToolVersion
+  -> m Bool
+guessTargetForSetCmd runLogger = \case
+  (SetGHCVersion GHCTargetVersion{..}) -> 
+    case T.toLower <$> _tvTarget of
+      Just target
+        | Just (target', other) <- findPrefix target
+          -- other may be `javascript-unknown-ghcjs`
+          -- hence the suggestion will be `ghcup set ghc javascript-unknown-ghcjs-9.6`
+        , let other' = if T.null other then "" else other <> "-"
+        -> errorWithSuggest (target' <> " " <> other' <> prettyVer _tvVersion)
+      Just target -> errorWithHelp $ "Unrecognized `" <> target <> "`."
+      Nothing
+        | Just (target', other) <- findPrefix (prettyVer _tvVersion)
+        -> errorWithSuggest (target' <> " " <> other)
+      Nothing -> warnOldStyle (prettyVer _tvVersion) -- Only has version
+  SetRecommended -> warnOldStyle "recommended"
+  SetNext -> warnOldStyle "next"
+  (SetToolTag tag) -> warnOldStyle $ T.pack $ tagToString tag
+  _ -> errorWithHelp "Impossible."
+  where
+  findPrefix target = case find (`T.isPrefixOf` target) ["ghc", "cabal", "hls", "stack"] of
+    Just prefix
+      | Just other <- T.stripPrefix prefix target
+      , let other' =
+              case T.uncons other of
+                Just ('-', rest) -> rest
+                _ -> other
+      -> pure (prefix, other')
+    _ -> Nothing
+
+  errorWithSuggest txt = do
+    runLogger (logError $ T.unlines ["Did you mean `ghcup set " <> txt <> "`?", "", "  View all candidates with `ghcup set --help`"])
+    pure False
+
+  -- There several valid options, so we print a warning and then continue
+  warnOldStyle txt = do
+    runLogger (logWarn $ "This is an old-style command for setting GHC. Use `ghcup set ghc " <> txt <> "` instead.")
+    pure True
+
+  errorWithHelp txt = do
+    runLogger (logError $ T.unlines [txt, "", "  View all candidates with `ghcup set --help`"])
+    pure False
 
 set :: forall m env.
        ( Monad m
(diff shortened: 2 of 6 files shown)


hasufell:
That looks interesting. I may be able to do proper review during weekend.

hasufell:
So, I think we might not need this.

If you look at the current branch [ghc-compile](https://github.com/haskell/ghcup-hs/tree/ghc-compile) from this PR: https://github.com/haskell/ghcup-hs/pull/854

Then you will see the behavior is already a bit different:

```
ghcup install ghc-9.2.5  3m  Thu 11:56
[ Warn  ] This is an old-style command for installing GHC. Use 'ghcup install ghc' instead.
[ Info  ] downloading: https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-0.0.7.yaml as file /home/hasufell/.ghcup/cache/ghcup-0.0.7.yaml
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[ Error ] [GHCup-00010] Unable to find a download for the requested version/distro.
[ Error ] Also check the logs in /home/hasufell/.ghcup/logs
```

This makes sense, because now we have proper cross support. So what happens here is that `ghc-9.2.5` gets parsed as:

```hs
GHCTargetVersion { _tvTarget = Just "ghc", _tvVersion = 9.2.5 }
```

Previously, it would throw away the `ghc` part. Now it doesn't anymore, so it becomes a fatal error.

What we want to do here now is to just give a better error description.

The error that's thrown and that triggers this message is here:

https://github.com/haskell/ghcup-hs/blob/6623e4b1c8804adf47aee244aed17e74abf02cf1/lib/GHCup/Errors.hs#L208-L218

We throw it in several places, e.g.:

https://github.com/haskell/ghcup-hs/blob/6623e4b1c8804adf47aee244aed17e74abf02cf1/lib/GHCup/Download.hs#L319-L326

So we might just have to add more info to the `NoDownload` type (the `GHCTargetVersion` for instance) and then check if `_tvTarget` matches any of the existing tools (ghc/cabal/hls/stack) and then be more descriptive in the error.

That way we have pushed down this concern out of the parser.

I quite like the optparse test suite though. I think this should be extended.

hasufell:
On second thought... I think my proposal has two issues:

* we lose information about the exact command that was run, so the message will look something like `"Perhaps you meant: 'ghcup <command> ghc 9.2.5'"`
* the warning will also trigger for `ghcup install ghc ghc-9.2.5`, in which case it will be somewhat confusing

hasufell:
Maybe it isn't too bad:

```
$ ghcup install ghc-9.2.5
[ Warn  ] This is an old-style command for installing GHC. Use 'ghcup install ghc' instead.
[ Error ] [GHCup-00010] Unable to find a download for ghc-9.2.5
[ ...   ] Perhaps you meant: 'ghcup <command> ghc 9.2.5'
[ Error ] Also check the logs in /home/hasufell/.ghcup/logs

$ ghcup install ghc ghc-9.2.5
[ Error ] [GHCup-00010] Unable to find a download for ghc-9.2.5
[ ...   ] Perhaps you meant: 'ghcup <command> ghc 9.2.5'
[ Error ] Also check the logs in /home/hasufell/.ghcup/logs
```

hasufell:
https://github.com/haskell/ghcup-hs/commit/b46537fd005222fc63cd80fc1ba7a174db4cc5fb on branch https://github.com/haskell/ghcup-hs/compare/ghc-compile

July541:
> Maybe it isn't too bad:
> 
> ```
> $ ghcup install ghc-9.2.5
> [ Warn  ] This is an old-style command for installing GHC. Use 'ghcup install ghc' instead.
> [ Error ] [GHCup-00010] Unable to find a download for ghc-9.2.5
> [ ...   ] Perhaps you meant: 'ghcup <command> ghc 9.2.5'
> [ Error ] Also check the logs in /home/hasufell/.ghcup/logs
> 
> $ ghcup install ghc ghc-9.2.5
> [ Error ] [GHCup-00010] Unable to find a download for ghc-9.2.5
> [ ...   ] Perhaps you meant: 'ghcup <command> ghc 9.2.5'
> [ Error ] Also check the logs in /home/hasufell/.ghcup/logs
> ```

Not bad indeed, turns out that while I typed `ghcup set ghc-9.2` and got an error, I quickly ask `ghcup set -h` for help. So it may not be worth to introduce much change, especially I'm not totally familiar with this system.

The only nit is `Perhaps you meant: 'ghcup <command> ghc 9.2.5'`, the `<command>` a little vague.

July541:
> I quite like the optparse test suite though. I think this should be extended.

I'll try to write some test for optparse on a new thread.

But it seems hard to take coverage test since the parser and the functionality are tangled.

July541:
> I'm not sure it's worth it.

I'm in favor of the current solution.

hasufell:
Some more tweaking leaves us with this:

```
ghcup install ghc-9.2.5
[ Warn  ] This is an old-style command for installing GHC. Use 'ghcup install ghc' instead.
[ Error ] [GHCup-00010] Unable to find a download for GHC version 'ghc-9.2.5' on detected platform x86_64-linux-fedora-37
[ ...   ] Perhaps you meant: 'ghcup <command> ghc 9.2.5'
[ Error ] Also check the logs in /home/hasufell/.ghcup/logs
```

I think it's quite useful to see what platform GHCup came up with when resolving the downloads.

July541:
Meaningful improvement!

Think from scratch, the error message `Unable to find a download for ghc-9.2.5`  is not correct, it turns out the failure reason is `Unable to find a download for ghc-ghc-9.2.5` right?

And furthermore, the user may think if `ghc-9.2.5` doesn't support x86_64-linux-fedora-37 or `ghc-9.2.5` doesn't exist at all... Do we have a manual that listed all candidate options, otherwise can we have this?

hasufell:
> Think from scratch, the error message Unable to find a download for ghc-9.2.5 is not correct, it turns out the failure reason is Unable to find a download for ghc-ghc-9.2.5 right?

That's what the new error message basically says:

```
Unable to find a download for GHC version 'ghc-9.2.5'
```

> Do we have a manual that listed all candidate options, otherwise can we have this?

`ghcup list`?

July541:
> ghcup list?

Well, I'm looking at `ghcup set` command and found there are many candidates that are not listed, likes `Tag` and `Day`

https://github.com/haskell/ghcup-hs/blob/d143daeb9a021ac7bd9c1fc1999a057ebc673547/lib/GHCup/Types.hs#L709-L712
https://github.com/haskell/ghcup-hs/blob/d143daeb9a021ac7bd9c1fc1999a057ebc673547/lib/GHCup/Types.hs#L156-L165

hasufell:
I'm not sure I understand what you mean by candidates.

PR #865, opened 2023-07-22
Add codec and schema for `GHCupInfo`

### LINKED ISSUE #90: Create yaml schema
In GitLab by @maerwald on Oct 19, 2020, 18:15

tfausak:
Fixes #90. 

It's not the prettiest schema, but it is _a_ schema. 

<details>
<summary>Click to expand output of <code>Autodocodec.Schema.jsonSchemaViaCodec @GHCup.Types.GHCupInfo</code>.</summary>

``` json
{
  "$defs": {
    "GHCupInfo": {
      "$comment": "GHCupInfo",
      "properties": {
        "ghcupDownloads": {
          "additionalProperties": {
            "additionalProperties": {
              "$defs": {
                "VersionInfo": {
                  "$comment": "VersionInfo",
                  "properties": {
                    "arch": {
                      "additionalProperties": {
                        "additionalProperties": {
                          "additionalProperties": {
                            "$ref": "#/$defs/DownloadInfo"
                          },
                          "type": "object"
                        },
                        "type": "object"
                      },
                      "type": "object"
                    },
                    "changeLog": {
                      "$defs": {
                        "URI": {
                          "type": "string"
                        }
                      },
                      "$ref": "#/$defs/URI"
                    },
                    "postInstall": {
                      "type": "string"
                    },
                    "postRemove": {
                      "type": "string"
                    },
                    "preCompile": {
                      "type": "string"
                    },
                    "releaseDay": {
                      "$comment": "Day"
                    },
                    "sourceDL": {
                      "$defs": {
                        "DownloadInfo": {
                          "$comment": "DownloadInfo",
                          "properties": {
                            "cSize": {
                              "type": "number"
                            },
                            "hash": {
                              "type": "string"
                            },
                            "output": {
                              "type": "string"
                            },
                            "subdir": {
                              "$defs": {
                                "TarDir": {
                                  "oneOf": [
                                    {
                                      "type": "string"
                                    },
                                    {
                                      "$comment": "RegexDir",
                                      "properties": {
                                        "RegexDir": {
                                          "type": "string"
                                        }
                                      },
                                      "required": [
                                        "RegexDir"
                                      ],
                                      "type": "object"
                                    }
                                  ]
                                }
                              },
                              "$ref": "#/$defs/TarDir"
                            },
                            "uri": {
                              "$ref": "#/$defs/URI"
                            }
                          },
                          "required": [
                            "hash",
                            "uri"
                          ],
                          "type": "object"
                        }
                      },
                      "$ref": "#/$defs/DownloadInfo"
                    },
                    "tags": {
                      "items": {
                        "$defs": {
                          "Tag": {
                            "type": "string"
                          }
                        },
                        "$ref": "#/$defs/Tag"
                      },
                      "type": "array"
                    },
                    "testDL": {
                      "$ref": "#/$defs/DownloadInfo"
                    }
                  },
                  "required": [
                    "arch",
                    "tags"
                  ],
                  "type": "object"
                }
              },
              "$ref": "#/$defs/VersionInfo"
            },
            "type": "object"
          },
          "type": "object"
        },
        "globalTools": {
          "additionalProperties": {
            "$ref": "#/$defs/DownloadInfo"
          },
          "type": "object"
        },
        "toolRequirements": {
          "additionalProperties": {
            "additionalProperties": {
              "additionalProperties": {
                "additionalProperties": {
                  "$defs": {
                    "Requirements": {
                      "$comment": "Requirements",
                      "properties": {
                        "distroPKGs": {
                          "items": {
                            "type": "string"
                          },
                          "type": "array"
                        },
                        "notes": {
                          "type": "string"
                        }
                      },
                      "required": [
                        "notes",
                        "distroPKGs"
                      ],
                      "type": "object"
                    }
                  },
                  "$ref": "#/$defs/Requirements"
                },
                "type": "object"
              },
              "type": "object"
            },
            "type": "object"
          },
          "type": "object"
        }
      },
      "required": [
        "globalTools",
        "ghcupDownloads",
        "toolRequirements"
      ],
      "type": "object"
    }
  },
  "$ref": "#/$defs/GHCupInfo"
}
```

</details>

I'm happy to make whatever changes. This was a pretty rough cut. I just added `HasCodec GHCupInfo` and kept adding instances until it compiled.

### DIFF
--- a/ghcup.cabal
+++ b/ghcup.cabal
@@ -111,6 +111,8 @@ library
   build-depends:
     , aeson                 >=1.4
     , async                 >=0.8        && <2.3
+    , autodocodec           ^>=0.2.0.3
+    , autodocodec-schema    ^>=0.1.0.3
     , base                  >=4.12       && <5
     , base16-bytestring     >=0.1.1.6    && <1.1
     , binary                ^>=0.8.6.0
@@ -140,6 +142,7 @@ library
     , retry                 ^>=0.8.1.2 || ^>=0.9
     , safe                  ^>=0.3.18
     , safe-exceptions       ^>=0.1
+    , scientific            ^>=0.3.7.0
     , split                 ^>=0.2.3.4
     , streamly              ^>=0.8.2
     , strict-base           ^>=0.4
--- a/lib/GHCup/Types/JSON.hs
+++ b/lib/GHCup/Types/JSON.hs
@@ -2,11 +2,13 @@
 
 {-# LANGUAGE DataKinds             #-}
 {-# LANGUAGE DeriveGeneric         #-}
+{-# LANGUAGE DerivingVia           #-}
 {-# LANGUAGE FlexibleContexts      #-}
 {-# LANGUAGE FlexibleInstances     #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings     #-}
 {-# LANGUAGE QuasiQuotes           #-}
+{-# LANGUAGE StandaloneDeriving    #-}
 {-# LANGUAGE TemplateHaskell       #-}
 {-# LANGUAGE TypeFamilies          #-}
 
@@ -29,6 +31,7 @@ import           Control.Applicative            ( (<|>) )
 import           Data.Aeson              hiding (Key)
 import           Data.Aeson.TH
 import           Data.Aeson.Types        hiding (Key)
+import           Data.Bifunctor
 import           Data.ByteString                ( ByteString )
 import           Data.List.NonEmpty             ( NonEmpty(..) )
 import           Data.Text.Encoding            as E
@@ -37,7 +40,9 @@ import           Data.Void
 import           URI.ByteString
 import           Text.Casing
 
+import qualified Autodocodec                   as Codec
 import qualified Data.List.NonEmpty            as NE
+import qualified Data.Scientific               as Scientific
 import qualified Data.Text                     as T
 import qualified Data.Text.Encoding.Error      as E
 import qualified Text.Megaparsec               as MP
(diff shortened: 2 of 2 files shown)


ghcup.cabal
```diff
@@ -111,6 +111,8 @@ library
   build-depends:
     , aeson                 >=1.4
     , async                 >=0.8        && <2.3
+    , autodocodec           ^>=0.2.0.3
```
hasufell, line 114: I'm **very** mindful about the dependency footprint of GHCup. It doesn't seem we need or want this dependency for reasons other than generating the schema? Can we do with CPP and a flag?

Also... we should have a binary that generates the actual schema. Preferably in ghcup-metadata repo: https://github.com/haskell/ghcup-metadata/tree/develop/ghcup-gen

tfausak: Autodocodec also provides the `FromJSON` and `ToJSON` instances, although obviously you can get those with Aeson alone. So in that sense it is just for the schema. 

You could hide the dependency behind a flag, but at that point I think you'd be better off simply adding a `schema.json` file to the repo somewhere. The benefit of Autodocodec is making sure that the instances all agree with each other. If it's only used conditionally, you lose that benefit. 

I also try to avoid heavy dependencies. I think the only thing that Autodocodec adds beyond Aeson's dependencies is [the `validity` package](https://hackage.haskell.org/package/validity-0.12.0.1). 

`autodocodec-schema`, on the other hand, could certainly be hidden away behind a flag. Its only purpose is to turn the `Codec` into an actual schema. Perhaps it could be part of the test suite, as a golden test?

hasufell: I'm not a big believer in autogenerating schemas like people do with servant and swagger.

A schema is a user facing document that needs to be designed. Then we want to make sure our metadata files always adhere to it.

So:

- we have a metadata file
- ghcup can parse it (already checked)
- we have a schema
- our metadata files match the schema at all times (needs to be added to CI)

I'd also say we want to turn the json schema into yaml, so we can add comments etc.

At that stage, we won't really need any generation mechanism, I guess.

tfausak: In that case, I'll close this PR. I posted the generated schema in the description — hopefully it can be used as a starting point. 

I've used [AJV](https://ajv.js.org) for validating schemas in CI and it's worked well for me.

PR #868, opened 2023-07-30
pretty-print instances for VersionRange and VersionCmp types

arjunkathuria:
Adds `Pretty` type-class instances for `VersionRange` and `VersionCmp` types.

PR in response to #777 .

`ghcup-metadata` repo would use the `Pretty` instances to print out distro versions and ranges.


hasufell:
Merged

PR #869, opened 2023-08-05
Fix libffi

maxmosk:
In Debian (current version is 12) libffi6 is extrimaly old and was replaced by libffi8 (see https://packages.debian.org/en/source/bookworm/libffi). May be, better way is replace it by `libffi`, that included in each version of Debian distro.


hasufell:
Duplicate of https://github.com/haskell/ghcup-hs/issues/777

A more strategic fix is being implemented by @arjunkathuria

Issue #874, opened 2023-08-26
GHCup does not validate gpg sig if using a `file://` yaml url


hasufell:

hasufell:
Merged manually

PR #881, opened 2023-09-02
Validate gpg sig even if using file:// yaml url

hasufell:
Fixes https://github.com/haskell/ghcup-hs/issues/874


hasufell:
Merged manually

Issue #886, opened 2023-09-08
xattr failing due to unrecognized options (MacOS Ventura, M1)


amordahl:
Hi,

When trying to use GHCUp on MacOS 13.5.1, on an M1 Mac, I get the following error in my log:

```
[ Warn  ] [GHCup-00841] Process "xattr" with arguments ["-r", "-d", "com.apple.quarantine",
[ ...   ]                                 "/Users/austin/.ghcup/tmp/ghcup-859ec65efd29c2b7"] failed with exit code 64.
[ Info  ] Installing HLS
option -r not recognized

usage: xattr [-slz] file [file ...]
       xattr -p [-slz] attr_name file [file ...]
       xattr -w [-sz] attr_name attr_value file [file ...]
       xattr -d [-s] attr_name file [file ...]
       xattr -c [-s] file [file ...]

The first form lists the names of all xattrs on the given file(s).
The second form (-p) prints the value of the xattr attr_name.
The third form (-w) sets the value of the xattr attr_name to attr_value.
The fourth form (-d) deletes the xattr attr_name.
The fifth form (-c) deletes (clears) all xattrs.

options:
  -h: print this help
  -s: act on symbolic links themselves rather than their targets
  -l: print long format (attr_name: attr_value)
  -z: compress or decompress (if compressed) attribute value in zip format
```

I know that MacOS ships with alternative versions of common CLI utilities that often don't have the same command line options as their GNU counterparts; is this a similar situation? I can't find any resources on installing an alternative version of xattr online. Furthermore, is this going to negatively impact my installation, or is this warning safely ignorable? Thanks for any time spent helping me on this issue.

romac:
I ran into the same issue, and I suspect that GHCup uses whatever `xattr` is in `PATH`, in my case it was the one provided by Homebrew at `/opt/homebrew/bin/xattr`, which does not support the `-r` option.

Running the same command with `/usr/bin/xattr` worked for me.

Perhaps GHCup should explicitly invoke `/usr/bin/xattr` otherwise this will break for every Homebrew user.

Alternatively, we could push for Homebrew to add support for `-r` to their version of `xattr`.

PR #893, opened 2023-10-03
Use absolute path to `/usr/bin/xattr` instead of pulling whatever is in `PATH`

romac:
Closes: #886 

On macOS systems with the [xattr](https://pypi.org/project/xattr/) Python package installed, the latter will often take precedence over the system `xattr` at `/usr/bin/xattr`. The problem is that the Python version does not support the `-r` flag to act recursively over a directory.

This commit changes the invocation of `xattr` to use the absolute path to the system version of `xattr` at `/usr/bin/xattr`.

See a similar issue for `exiftool` [here](https://github.com/exiftool/exiftool/issues/122).

### DIFF
--- a/lib/GHCup/Utils.hs
+++ b/lib/GHCup/Utils.hs
@@ -1033,7 +1033,7 @@ applyPatches pdir ddir = do
 
   patches <- liftIO $ quilt `catchIO` (\e ->
     if isDoesNotExistError e || isPermissionError e then
-      lexicographical 
+      lexicographical
     else throwIO e)
   forM_ patches $ \patch' -> applyPatch patch' ddir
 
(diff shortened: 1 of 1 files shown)


hasufell:
Should we fall back to any PATH xattr if there is none in `/usr/bin`?

romac:
I am not sure it's even possible for it to not exist at `/usr/bin/xattr` given that `/usr/bin/` is protected by SIP and therefore read-only since macOS Mojave (10.14) at the very least. See these two articles for more info:

- https://eclecticlight.co/2019/08/29/macos-mojave-boot-volume-layout/
- https://eclecticlight.co/2020/01/23/catalina-boot-volumes/

hasufell:
Merged manually

PR #895, opened 2023-10-04
Update install.md: Debian 12 (bookworm) has libffi8, no libffi7

mfhepp:
In Debian 12 (bookworm), there is only libffi8, not libffi7

See https://packages.debian.org/search?suite=bookworm&keywords=libffi


hasufell:
The file you're editing is auto generated. 

See https://github.com/haskell/ghcup-metadata/blob/374a68e0746d65077d5e41bc8fe362b9b1a7a6d5/ghcup-0.0.7.yaml#L9

mfhepp:
So what would be the proper place to fix it? For Debian Bookwork, the resulting doc is simply incorrect ;-). Sorry - Haskell is pretty far from my core skills and work, just want to be helpful ...

hasufell:
> So what would be the proper place to fix it? 

The file I just linked above

PR #940, opened 2023-11-28
Update WASM install instructions

konsumlamm:
I tried to install the cross bindists for the JS and WASM backends on a GitHub action. Everything worked fine, except for `sh setup.sh`, which produced the following error:
```sh
setup.sh: 3: set: Illegal option -o pipefail
```
Changing the line to `./setup.sh` fixes the issue.


hasufell:
https://www.haskell.org/ghcup/guide/#ghc-wasm-cross-bindists-experimental

Issue #875, opened 2023-08-26
ghcup tui shows docs on ctrl-c


alanz:
When I am in `ghcup tui`, I instinctively hit `ctrl-c` to quit.

And I am always surprised to be taken to a web browser opened to the docs of whatever my cursor is on at the time.

The legend at the bottom does not mention this possibility, perhaps it should, bound to something sane like `d`

hasufell:
Changing the key mapping will break lots of users workflows. You can change it yourself: https://github.com/haskell/ghcup-hs/blob/d759535faaa0f7a75414e46ccb61e1ee2e5d4563/data/config.yaml#L33

We could also detect control keys and not trigger any of the actions if they are in combination with such.

alanz:
Ah, so it is looking for the key, regardless of modifiers. Can you make it ignore them if ctrl is pressed at the same time.

Going to the changelog is great, I just have a problem with <ctrl-c> doing it.

hasufell:
The code is here: https://github.com/haskell/ghcup-hs/blob/d759535faaa0f7a75414e46ccb61e1ee2e5d4563/app/ghcup/BrickMain.hs#L315

PRs welcome.

hasufell:
> The code is here:
> 
> https://github.com/haskell/ghcup-hs/blob/d759535faaa0f7a75414e46ccb61e1ee2e5d4563/app/ghcup/BrickMain.hs#L315
> 
> PRs welcome.

Basically, we're ignoring the modifier: https://hackage.haskell.org/package/vty-5.38/docs/Graphics-Vty-Input-Events.html#t:Event

https://github.com/haskell/ghcup-hs/blob/d759535faaa0f7a75414e46ccb61e1ee2e5d4563/app/ghcup/BrickMain.hs#L328-L331

Should be something like

```hs
    (VtyEvent (Vty.EvKey key [])) ->
      case find (\(key', _, _) -> key' == key) (keyHandlers kb) of
        Nothing -> put st
        Just (_, _, handler) -> handler st
```

alanz:
I decided I could live with it when I got

```
[15 of 15] Compiling Codec.Archive    ( src/Codec/Archive.hs, dist/build/Codec/Archive.o, dist/build/Codec/Archive.dyn_o )

c/archive_random.c:54:13: error:
     error: static declaration of ‘arc4random_buf’ follows non-static declaration
       54 | static void arc4random_buf(void *, size_t);
[...]

Issue #846, opened 2023-07-07
Improve 'ghcup compile ghc' interface


hasufell:
The main issues are:

1. passing additional flags to build configure... 
2. but also passing additional flags to bindist configure
3. and also passing additional flags to hadrian itself

Currently we support the first case via `--` longopts. But that can only cover one use case.

This is particularly important for cross builds, which may need different than default settings.

@hsyl20

hasufell:
Should probably also make hadrian default or detect whether a make build system exists. `--hadrian` flag should only be required if there's both make and hadrian.

chreekat:
> Should probably also make hadrian default or detect whether a make build system exists. `--hadrian` flag should only be required if there's both make and hadrian.

Is it possible to say "Use make for <9.6 and Hadrian for >= 9.6"?

And fwiw I think 9.4 was the only series that officially supported both, although I also think Hadrian was pretty immature then.

hasufell:
> Is it possible to say "Use make for <9.6 and Hadrian for >= 9.6"?

Currently, my idea is just to look for the hadrian files and if they're not there, assume make:

https://github.com/haskell/ghcup-hs/pull/854/files#diff-3b9b75e85b8e396244d24bf04ef933cae33b4d20df6f5889304bfb36696c43d6R957

The choice can be made explicit with `--hadrian | --make`.

chreekat:
I don't know that it's valuable to make it an option to users if 9.4 is the only version that officially supported both. <9.4, --hadrian might work but will be buggy (and the bugs will never get fixed); >=9.6, --make will fail.

hasufell:
Yes, but GHC has made it extra hard to detect which version you're actually compiling: https://gitlab.haskell.org/ghc/ghc/-/issues/22322

The fix was backported, but I have no clear overview at the moment what versions work and which don't. 

As such I'm moving away from logic that depends on version numbers. If you try to compile 9.6.1 with make, that may fail. But maybe not, because you're using custom patched bindists, a different branch etc (remember you can already supply patches before the build... that's explicitly supported).

Dynamic detection seems more flexible. 

What we probably want is a more descriptive error in case you try to build 9.6.1 with make: "The make build system has been removed since 9.6. You probably want to omit '--make'."

hasufell:
And I'll probably support the `VERSION` file, in case it exists.

hasufell:
What's missing now is the question on how to support additional arguments to `./configure` and `hadrian ...`.

[...]

Issue #706, opened 2022-12-07
Feature request: support for compiling HLS from ghcup tui?


david-christiansen:
The ability of GHCup to compile HLS from source for a given HLS and compiler version is an under-appreciated, valuable feature. I think that many users aren't aware of how easy this actually is.

Would it be feasible to add some kind of indication to `ghcup tui` that a given GHC that doesn't have an HLS binary can be used to attempt to compile it?

hasufell:
Feature wise this makes sense. It's just that it's not trivial to design. We already have a lot of hotkeys, so I don't want to overload it with more. This would mean to implement some type of menu or navigation.

So maybe:

- default is still opening the tools page
- we add one more hotkey `m` for opening the menu
- the menu will be a new page with 2 items 'tool bindists' and 'compile HLS'
- a command line option can be passed to `ghcup tui` to open a different page by default
- maybe this requires a more concise presentation of the hotkeys... only the most important should be shown by default and `?` should open a help page

Alternatively, a side pane menu could be possible.

arjunkathuria:
Hi! @david-christiansen feature discoverability is definitely an area we need to work on. As mentioned, not many people know about features like `ghcup compile`  or other useful features like isolated installs, which i would term as somewhat advanced features (i.e. not needed by most users in most cases, but pretty handy in their use-cases).

I don't think nested menus via a TUI is an optimal way to go here, we should keep the TUI simple.

I Think better upfront documentation, with a dedicated advanced section (something similar to what stack does with it's docs), would be more effective.

We can also think about adding prompts with helpful suggestions and hints where possible without them becoming tedious, overbearing or annoying, while keeping the quick _"come, install, exit"_ feel of the tool at the center of user experience.

hasufell:
@arjunkathuria that's an interesting stance.

I think it's true some of the success of GHCup might be due to the simplicity and that the tool gets out of your way.

So what is really the value of the TUI? Yeah, I think it's mostly a better presentation of the list of available and installed tools. So it's about concise information that can be easily navigated. Maybe it's much less about the "interactivity" of installation. Then the question would be what is the value of doing compilation in the TUI when it doesn't add to the value of information presentation? Filling in boxes and inputs instead of typing it in the cli?

I also have to note that GHCup has quite sophisticated shell completion, e.g. it will complete available HLS versions and all sorts of other things when you do `ghcup compile hls...`.

But maybe there's another reason why TUI for this might improve user experience?

[...]

PR #949, opened 2023-12-11
Context Menu

lsmor:
Create popUp widget: widget state, widget drawing and widget handler. The widget is pretty straigthfoward

I have to implement the instalation logic still. I am gonna need some help with it. I understand how to pass some advance parameters, but not others. The whole list of parameters is:

- url
- set
- isolate
- force
- configure Args

If we translate that into code

```haskell
--                                           |- Change be IsolateDir ... if parameter isolate is set
--                                           |             |- change to True is param force is set.
installGHCBin (GHCTargetVersion lCross lVer) GHCupInternal False [] $> (vi, dirs, ce)
--                                                               |- Is this list for the rest of params??
```


hasufell:
Nice.

Although it seems this popup is for installation? We want it for `ghcup compile hls`.

lsmor:
> Although it seems this popup is for installation? We want it for `ghcup compile hls`.

Ouch! I got confuse in the conversation. If the new feature is for `ghcup compile`, how would the user interface be? Another hot-key?

hasufell:
> > Although it seems this popup is for installation? We want it for `ghcup compile hls`.
> 
> Ouch! I got confuse in the conversation. If the new feature is for `ghcup compile`, how would the user interface be? Another hot-key?

I think we're going to run out of hotkeys.

I suggest something like "context menu", maybe defaulting to `Enter`. Then it has select menu which we may extend in the future.

lsmor:
> I suggest something like "context menu", maybe defaulting to `Enter`. Then it has select menu which we may extend in the future.

I think I am missing some `ghcup` knowledge here. From my understanding `ghcup compile` is a command like `ghcup install`, or `ghcup set`. So, Is the "context menu" in fact a "compile menu"? I mean, instead of adding a hot-key for compile, we have a popup with a big "compile" button and a buch of options (following `ghcup compile hls --help` docs). Maybe there is something about `compile` I am missing...

For reference I've tried to implement [this](https://github.com/haskell/ghcup-hs/issues/706#issuecomment-1356405427) comment ideas.

hasufell:
> So, Is the "context menu" in fact a "compile menu"? 

Let's say the context menu for now will have only one button that leads you to said compile menu.

We can add more stuff as we go.

lsmor:
> We can add more stuff as we go.

I am playing around with this. Would it be useful to add an small Menu widget utilities? It would be some default widgets for easily creating/composing menus. I am thinking three options `data MenuEntry = Button | EditBox | CheckBox`. Then you defined your menu as

```haskell
-- oversimplification
data CompileMenu =  CompileMenu {gitRefEditBox :: MenuEntry, setCheckBox :: MenuEntry, okButton :: MenuEntry}
newCompileMenu = CompileMenu EditBox CheckBox Button
```

hasufell:
Yeah, why not. We can always change things. Code decisions here are low impact, because they can be easily reverted.

It's harder with our decisions on how to organize installed files. They are hard to revert.

runeksvendsen:
Hi @lsmor

Thank you for your efforts in getting this implemented!

I want to help you get this over the finish line.

To achieve this, I think we first need a clear specification of exactly what it is we want to implement. I have read the linked issue (https://github.com/haskell/ghcup-hs/issues/706) as well as the comments in this PR, and below is my suggestion for a final spec (for this PR):

1. A new "context menu" is added
   1. This is a small popup widget (overlaying part of the main TUI widget) that appears when the user presses ENTER on the currently selected tool (GHCup/Stack/HLS/cabal/GHC) in the TUI
   2. The title of this popup widget is `Context menu (<tool name> <tool version>)`, for example `Context menu (GHC 8.10.7)`
   3. The widget contains one or more actions relevant to the given tool. Each action is a line that can be navigated to using the up/down keys and entered by pressing ENTER. All configuration for the given action happens in the widget that appears after pressing ENTER on an action (to keep the context menu clean).
1.  Generally speaking, this context menu will contain additional actions relevant to the selected tool in question. Thus, the content of the context menu may differ depending on which tool is selected. For example:
    1. For GHC only there may be a "compile HLS for this GHC version"-item
    2. For all tools:
        1. An "advanced installation" item, which allows configuring advanced options such as "isolated install"/"force install"/"compile instead of downloading bin-dist" (the options displayed by `install --help`)
        2. A "compile" item, which acts as a TUI for the `ghcup compile` command
    3. For HLS only we can imagine being able to deselect plugins

Example context menu for GHC 8.10.7:

```
┌────────────────────────────────────────GHCup────────────────────────────────────────┐
│    Tool  Version         Tags                          Notes                        │
│─────────────────────────────────────────────────────────────────────────────────────│
│                                                                                     │
│─────────────────────────────────────────────────────────────────────────────────────│
│                                                                                     │
│          ┌───────────────────Context menu (GHC 8.10.7)───────────────────┐          │
│          │┌────────┐                                                     │          │
│          ││Compile │                                                     │          │
│          │└────────┘                                                     │          │
│          │┌─────────────────────┐                                        │          │
│──────────││Advanced installation│                                        │──────────│
│          │└─────────────────────┘                                        │          │
│          │┌────────────────────────────────────────┐                     │          │
│          ││Compile HLS for this GHC version        │                     │          │
│          │└────────────────────────────────────────┘                     │          │
│          └───────────────────────────────────────────────────────────────┘          │
│                                                                                     │
│                                                                                     │
│─────────────────────────────────────────────────────────────────────────────────────│
│                                                                                     │
└─────────────────────────────────────────────────────────────────────────────────────┘
q:Quit  i:Install  u:Uninstall  s:Set  c:ChangeLog  a:Show all versions  ↑:Up  ↓:Down
h:help
```

Note that the above is only about the nature of the "context menu", which allows putting more advanced features into the TUI without cluttering the main view and adding additional hot keys.

Does this sound reasonable? @hasufell @arjunkathuria @david-christiansen @Kleidukos

If we agree on this, then it's up to you @lsmor to decide which of the above context menu features you want to implement. It looks like you've started to implement the "advanced installation"-item feature. The goal of this PR can then be adding (1) the context menu itself; and (2) whatever context menu feature @lsmor wants to implement.

If @lsmor wants to continue with the "advanced installation"-item feature then a subsequent PR can add the "compile HLS for this GHC version"-feature.

lsmor:
> Code decisions here are low impact, because they can be easily reverted.

Well, given the amount of features for this (I wont implement all of them in this PR though), I think It is worth to split the `brick` related stuff into another component (say `lib-tui`), instead of having a single file. So I am taking this quote literal, and do It. The only thing I wonder is if this is going to break CI for some reason. In principle I am not planning to break any cabal flag so I guess there should be no problem

> I want to help you get this over the finish line.

Thanks, actually I had somethings wrong. This design helped to get things clearer

runeksvendsen:
> Thanks, actually I had somethings wrong. This design helped to get things clearer

Great! I'm happy to hear that.

> I have to implement the instalation logic still. I am gonna need some help with it. I understand how to pass some advance parameters, but not others. The whole list of parameters is:
> 
> * url
> * set
> * isolate
> * force
> * configure Args
> 
> If we translate that into code
> 
> ```haskell
> --                                           |- Change be IsolateDir ... if parameter isolate is set
> --                                           |             |- change to True is param force is set.
> installGHCBin (GHCTargetVersion lCross lVer) GHCupInternal False [] $> (vi, dirs, ce)
> --                                                               |- Is this list for the rest of params??
> ```

I'm not that familiar with the GHCup codebase, but after looking around a bit, I can see that all the relevant options are contained in the data type `InstallOptions` (in the `GHCup.OptParse.Install` module).

So I would prefer to go a few levels up from `installGHCBin`, and calling `GHCup.OptParse.Install.install`. This takes an `InstallOptions` as argument and ultimately calls `installGHCBin`. This should also work for all tools instead of just GHC, which we need because the `Advanced installation` action should be in the Context Menu for all tools.

Generally speaking, I prefer to call some sort of `eval` function which takes a single data type as argument that specifies everything to do. This avoids having to call specific functions with specific arguments in the right place. In this case the `install` function seems to behave this way along with the `InstallCommand` data type.

@hasufell does the above sound reasonable to you? I'm not familiar with how various options are passed around by GHCup, so calling `GHCup.OptParse.Install.install` from an action within the Context menu might be problematic.

hasufell:
> @hasufell does the above sound reasonable to you? I'm not familiar with how various options are passed around by GHCup, so calling `GHCup.OptParse.Install.install` from an action within the Context menu might be problematic.

I'm not sure I agree. `GHCup.OptParse.Install.install` was written for the cli interface and assumes cli when doing error handling, logging, etc.

My approach in such cases, where it's not clear what the design overlap is, usually is to rather duplicate code and let us explore what actually ends up similar. Abstraction is, imo, discovered and not actively sought.

Here I'm not sure what a good design is that abstracts over cli and tui. A pre-existing interface may just skew your creativity when coming up with a good API for the tui to consume.

Also remember that this creates coupling and refactors over the optparse code would also affect the tui.

That said, I'd probably expect a small common denominator between cli and tui in a separate sublibrary or the main library, if one really wants to avoid code/logic duplication. So you might end up with 3 ADTs:

- TUI ADT
- opt-parse ADT
- GHCup library ADT, which the other two are converted to

And now I'm not sure it's worth it. But feel free to give it a try.

runeksvendsen:
@hasufell thank you for your input! You make several good points.

> And now I'm not sure it's worth it.

I agree. Let's go with the simple solution for now.

@lsmor I will try again to answer your question — please ignore my comment above :-)

> * url
> * set
> * isolate
> * force
> * configure Args
> 
> If we translate that into code
> 
> ```haskell
> --                                           |- Change be IsolateDir ... if parameter isolate is set
> --                                           |             |- change to True is param force is set.
> installGHCBin (GHCTargetVersion lCross lVer) GHCupInternal False [] $> (vi, dirs, ce)
> --                                                               |- Is this list for the rest of params??
> ```

As you've figured out, the parameters `isolate`, `force` and `configure args` are all arguments to `installGHCBin`. 

If the `url` option is set (the CLI option called `instBindist`), then we call `installGHCBindist` instead of `installGHCBin` as is done here: https://github.com/haskell/ghcup-hs/blob/df192ee18e64b6d080adc14f498d97853274a543/lib-opt/GHCup/OptParse/Install.hs#L342-L348 and we also run the `Excepts` actions with the `noVerify` setting set to `True`: https://github.com/haskell/ghcup-hs/blob/df192ee18e64b6d080adc14f498d97853274a543/lib-opt/GHCup/OptParse/Install.hs#L340

For the `set` option, the following line takes care of this: https://github.com/haskell/ghcup-hs/blob/df192ee18e64b6d080adc14f498d97853274a543/lib-opt/GHCup/OptParse/Install.hs#L349

Let me know whether this clears things up.

lsmor:
I am force pushing a commit with a big (but easy) change. I have moved the BrickMain.hs into its own library and split it in many modules. The original BrickMain.hs remains as in master branch. The reasons for this are:

- If we are willing to implement a few menus with advance options, one module becomes too much of a mess to handle
- If we gonna need more than one module, a library becomes a better approach (IMO) than many modules in an executable component
- (selfish reason) This wont be an easy PR, hence a lot of rebase / conflicts are expected. I am willing to re-implement changes in BrickMain.hs into the new library, but rebasing 10 commits each with many conflicts is just painfull (not a commit expert myself)

runeksvendsen:
@hasufell what do you think about merging this as-is (without the "context menu"-changes)? It works for me, and I think it's a useful change.

ghcup.cabal
```diff
@@ -281,6 +281,60 @@ library ghcup-optparse
...
+    -fwarn-incomplete-record-updates
+
+  build-depends:  
+    , ghcup
+    , ghcup-optparse
+    , optics        ^>=0.4
+    , brick         ^>=2.1
+    , transformers  ^>=0.5
+    , vty           ^>=6.0
+    , unix          ^>=2.7
```
runeksvendsen, line 333: should not be here `if os(windows)`

lsmor: I've push a fix. Actually the conditional was below those lines

lsmor:
Hi there!

I'll be pushing some funcitonality soon. For now I've implemented only the visual part, without logic. Before continuing I'd like to have some feedback. Each tools has its own "Context Menu" (I'd prefer advance options, honestly), from there you can go to other settings like "compile", "install", etc...

install-pop-up


Code is very messi, I am cleaning it a liitle bit before pushing.

hasufell:
@lsmor if you rebase this branch against master, then CI should succeed.

hasufell:
@lsmor let us know if you need any further feedback for the design. I believe this is a major improvement to the TUI.

lsmor:
I am pushing (only) the visuals. TODOS:

- [ ] migrate the fix #987 to the new library
- [ ] implement the visual for CompileMenu
- [ ] implement the logic for AdvanceInstallMenu
- [ ] implement the logic for CompileMenu

lsmor:
> @lsmor what's the timeline you think you will have this finished? I'm evaluating whether to wait for it before making a release or doing a release sooner than later.

um... I don't expect it to have it soon. the first two points are easy, and I can have them in the next week, but I still don't know how difficult will be to have the last two point working. + There should be some extensive testing which I don't know how to approach actually. So I would expect to have something testable in about 1,5 month.  From there, fixing all possible bugs, etc... let say 2 or 3 months to merge.

Btw, I am terrible at ETA. Notice #850 , took me 5+ months (summer in the middle) and I estimated 3 months (and it was merge with a bug on my side!!!)... Also I am more familiar with the code base now.

hasufell:
> I am terrible at ETA

Well, I'm not trying to be a manager :D

Just trying to understand when to schedule releases. I think I'll start setting up a milestone then and only merge smaller stuff.

lsmor:
> Maybe the same as in the main window, which by default is `q`

I just recall, why this isn't a good Idea. You can either type `q` on a textbox, or `q` to exit. You could be intelligent enough and dispatch the `q` depending on whether or not a text box is focused, but thats brittle as it literally depends on pattern matching order... Also, it feels a little bit clumsy. 

Adding a different key for exit, is risky as something like `ESC` might be dispatch to the terminal itself instead of the content within the terminal (example: `Ctrl + Enter` doesn't work on my gnome's terminal, probably because some colission with terminal's shortcuts)

¿is it ok if we keep the cancell button?

> I think it should be possible to put that help text inside the input fields

I am adding this to the check list

hasufell:
> I just recall, why this isn't a good Idea. You can either type q on a textbox, or q to exit. You could be intelligent enough and dispatch the q depending on whether or not a text box is focused, but thats brittle as it literally depends on pattern matching order... Also, it feels a little bit clumsy.

I'm not sure I follow. It's very normal that the semantics of key-presses changes depending on which window/box is focused. That's literally how window managers work.

We could make  `ctrl+c` be the default for "exit everything".

lsmor:
> It's very normal that the semantics of key-presses changes depending on which window/box is focused

Sure!, so it does in ghcup. Let me ilustrate the problem better.

```
┌────────────────────────────────────────GHCup────────────────────────────────────────┐
│    Tool  Version         Tags                          Notes                        │
│─────────────────────────────────────────────────────────────────────────────────────│
│                                                                                     │
│─────────────────────────────────────────────────────────────────────────────────────│
│                                                                                     │
│          ┌───────────────────Compile (GHC 8.10.7)────────────────────────┐          │
│          │┌────────┐             ┌────────────────────────────────┐      │          │
│          ││isolate │             │                                │      │          │
│          │└────────┘             └────────────────────────────────┘      │          │
│          │┌────┐                 ┌─────┐                                 │          │
│──────────││set │                 │  X  │                                 │──────────│
│          │└────┘                 └─────┘                                 │          │
│          │ .                                                             │          │
│          │ .                                                             │          │
│          │ .                                                             │          │
│          └───────────────────────────────────────────────────────────────┘          │
│                                                                                     │
│                                                                                     │
│─────────────────────────────────────────────────────────────────────────────────────│
│                                                                                     │
└─────────────────────────────────────────────────────────────────────────────────────┘
q:Quit  i:Install  u:Uninstall  s:Set  c:ChangeLog  a:Show all versions  ↑:Up  ↓:Down
h:help
```
In the example above arrow keys are dispatched to the upfront layer, which is the focused on. That's the expected behaviour. The problem comes when we are over `isolate` field and we press `q`; it should write q on the text box, but, If we are over `set` field, it should exit the menu. Aside from the code complexity and brittleness of it, the user experience is a bit unintuitive.

As for now, I'll continue to implement the missing features (this isn't an stopper). There is a known bug that `q` exits even if we are over a textbox. I'll add that bug to the checklist

lsmor:
Rebased + Help messages + migration of #987. I think help messages do look cool. Slow but steady progress.

Peek 2024-02-07 22-34

lsmor:
Small update. It turns out that the API I created for Menus is very very bad. In particular, it can't validate fields with good error messages, and also is very unsafe and error prone. That led to very buggy code, and eventualy I decided to refactor the whole Menu API. I am using Brick.Form as the inspiration for the new API.

lsmor:
(unrelated disclaimer: I don't know what's going on with CI. It builds in my local computer...)

Hi there. 

I am stuck with the `ghcup` code base. I am implementing the Advance install functionality. I think I am running the same code (or equivalent code) as in `ghcup-optparse`, but I am getting a weird error.

to reproduce:

- Install a tool isolated via cli. ex: `ghcup install stack 2.15.1 --isolated /some/path/to/test`
- `stack` binary should be in that directory. Delete it.
- Do the same via tui.
  - go to the same tool 
  - press `Enter`
  - Open the install Menu
  - fill the isolated field with `/isolated/some/path/to/test`
  - press `Enter` on the button on the top

Interestingly the same log is outputed, but it fails with

```
Error: [GHCup-00070] Unable to copy a file. Reason was: /path/to/test
/stack: openFd: already exists (File exists)
```

I don't think this is due to installing via cli first, as I can run `ghcup install stack 2.15.1 --isolated /some/path/to/test
` multiple times as long as I delete `/some/path/to/test/stack` before.

@hasufell do you have any intuition about this?


For completness. Here are both outputs

```bash
> ghcup install stack 2.15.1 --isolate /home/luis/delete
[ Info  ] downloading: https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-0.0.8.yaml as file /home/luis/.ghcup/cache/ghcup-0.0.8.yaml
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[ Warn  ] New ghc version available. If you want to install this latest version, run 'ghcup install ghc 9.8.2'
[ Info  ] downloading: https://downloads.haskell.org/~ghcup/unofficial-bindists/stack/2.15.1/stack-2.15.1-linux-x86_64.tar.gz as file /home/luis/.ghcup/tmp/ghcup-6148c2bcd44f02bb/stack-2.15.1-linux-x86_64.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 20.5M  100 20.5M    0     0  14.3M      0  0:00:01  0:00:01 --:--:-- 14.3M
[ Info  ] verifying digest of: stack-2.15.1-linux-x86_64.tar.gz
[ Info  ] Unpacking: stack-2.15.1-linux-x86_64.tar.gz to /home/luis/.ghcup/tmp/ghcup-20341ce62eefc284
[ Info  ] isolated installing Stack to /home/luis/delete
[ Info  ] Installing stack
[ Info  ] Stack installation successful
[ Info  ] Stack manages GHC versions internally by default. To improve integration, please visit:
[ ...   ]   https://www.haskell.org/ghcup/guide/#stack-integration
[ ...   ] 
[ ...   ] Also check out:
[ ...   ]   https://docs.haskellstack.org/en/stable/yaml_configuration
[ ...   ] 

> rm /home/luis/delete/stack

> cabal run ghcup -- tui
... # do whatever on tui

[ Info  ] downloading: https://downloads.haskell.org/~ghcup/unofficial-bindists/stack/2.15.1/stack-2.15.1-linux-x86_64.tar.gz as file /home/luis/.ghcup/tmp/ghcup-8dab1a41a66d3f3d/stack-2.15.1-linux-x86_64.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 20.5M  100 20.5M    0     0  14.7M      0  0:00:01  0:00:01 --:--:-- 14.7M
[ Info  ] verifying digest of: stack-2.15.1-linux-x86_64.tar.gz
[ Info  ] Unpacking: stack-2.15.1-linux-x86_64.tar.gz to /home/luis/.ghcup/tmp/ghcup-70ce6210c3c83de3
[ Info  ] isolated installing Stack to /home/luis/delete
[ Info  ] Installing stack
Error: [GHCup-00070] Unable to copy a file. Reason was: /home/luis/delete
/stack: openFd: already exists (File exists)
Also check the logs in ~/.ghcup/logs
Press enter to continue
```

hasufell:
Yes, the issue is that the text inbox causes a newline, so you're not installing into `/home/luis/delete` but into `/home/luis/delete\n`.

lsmor:
Finally this is ready for review. I think the UX is good, but I haven't tested the functionality very well... because I don't know how. Actually, the only option I've tested is the "set after install" option. Let me add a list of things to test / improve

- [ ] old functionality works as usual (can be tested asap. I've done it but second review is a must)
- [ ] every field in every menu has a well-behaving validator. Mostly I copy-paste validators from lib-opt, so an input is valid in the TUI iff is valid in the CLI. (you get instant feedback for wrong inputs in the TUI, so It can be reviewed asap without actually compiling nor installing a thing.).
- [ ] Tools are installed / compile as expected. (I don't know how to test this... Most options are greek to me)
- [ ] UX is intuitive. Particullary, exiting a Menu is a little bit clumsy, because exit key works only when we focus on a button
- [ ] Currently `ghcup-tui` depends on `ghcup-optparse`. This is not ideal (see [comment](https://github.com/haskell/ghcup-hs/pull/949#issuecomment-1869554009)), but is convinient for review since a lot of code is copy pasted between both libraries.  We have to decide either if _we are ok with duplication_ or _we are ok with dependencies_ or _would be better to create a separate component_. 
- [ ] CI is broken. What have I done?

Strangely enough, some code is duplicated and some is copy-pasted. The reason for this is just lazyness. I wanted to push a ready-to-review version asap, and then fix all broken things

lib-tui/GHCup/Brick/App.hs
```diff
@@ -0,0 +1,212 @@
...
+    VtyEvent (Vty.EvKey (Vty.KChar 'q') _ ) -> mode .= Navigation
+    _ -> pure ()
+
+-- | Tab/Arrows to navigate. 
+navigationHandler :: BrickEvent Name e -> EventM Name BrickState ()
+navigationHandler ev = do
+  AppState { keyBindings = kb } <- liftIO $ readIORef Actions.settings'
+  case ev of
+    inner_event@(VtyEvent (Vty.EvKey key _)) ->
+      case find (\(key', _, _) -> key' == KeyCombination key []) (Actions.keyHandlers kb) of
```
hasufell, line 122: This is a regression wrt mod behavior: https://github.com/haskell/ghcup-hs/issues/875

```suggestion
    inner_event@(VtyEvent (Vty.EvKey key mods)) ->
      case find (\(key', _, _) -> key' == KeyCombination key mods) (Actions.keyHandlers kb) of
```

hasufell:
Navigation feels a bit hard with `q` being ambiguous while you're on an input field. I'd say we should support `ctrl+c` as a shortcut to close whatever window/popup is currently active. This will also work with input fields.

lib-tui/GHCup/Brick/Widgets/Menus/CompileGHC.hs
```diff
@@ -0,0 +1,207 @@
...
+          & Menu.fieldHelpMsgL .~ "Set as active version after install"
+      , Menu.createEditableField (Common.MenuElement Common.OvewrwiteVerEditBox) versionV overwriteVer
+          & Menu.fieldLabelL .~ "overwrite-version"
+          & Menu.fieldHelpMsgL .~ "Allows to overwrite the finally installed VERSION with a different one"
+      , Menu.createEditableField (Common.MenuElement Common.BuildFlavourEditBox) (Right . Just . T.unpack) buildFlavour
+          & Menu.fieldLabelL .~ "flavour"
+          & Menu.fieldHelpMsgL .~ "Set the compile build flavour (this value depends on the build system type: 'make' vs 'hadrian')"
+      , Menu.createEditableField (Common.MenuElement Common.BuildSystemEditBox) systemV buildSystem
+          & Menu.fieldLabelL .~ "build system"
+          & Menu.fieldHelpMsgL .~ "Set the compile build flavour (this value depends on the build system type: 'make' vs 'hadrian')"
```
hasufell, line 189: This seems to not correspond to "build system", but "flavour"

lib-tui/GHCup/Brick/Widgets/Menus/CompileGHC.hs
```diff
@@ -0,0 +1,207 @@
...
+
+    additionalValidator :: T.Text -> Either Menu.ErrorMessage [T.Text]
+    additionalValidator = Right . T.split isSpace
+
+    systemV :: T.Text -> Either Menu.ErrorMessage (Maybe BuildSystem)
+    systemV = whenEmpty Nothing readSys
+      where
+        readSys i
+          | T.toLower i == "hadrian" = Right $ Just Hadrian
+          | T.toLower i == "make"    = Right $ Just Hadrian
```
hasufell, line 155: ?

hasufell: Also, the "build system" input field doesn't work for me. It always says "Not a valid Build System".

hasufell:
> Currently ghcup-tui depends on ghcup-optparse. This is not ideal (see https://github.com/haskell/ghcup-hs/pull/949#issuecomment-1869554009), but is convinient for review since a lot of code is copy pasted between both libraries. We have to decide either if we are ok with duplication or we are ok with dependencies or would be better to create a separate component.

So what is your intuition on it? What is the common abstraction between the two?

hasufell:
Btw., your editor leaves a lot of trailing spaces, which show up in my editor. Can you configure your editor to not save trailing whitespace?

lib-tui/GHCup/Brick/Widgets/Menus/CompileGHC.hs
lsmor: I guess you get the "not valid" because Trailing new line from Brick's editor. I'll add a generic "drop traling newline"  when validating, because I've seen this a few times already

lib-tui/GHCup/Brick/App.hs
lsmor: yes. I guess I messed with the rebase, sorry. I am changing it

lsmor:
DISCLAIMER: The whole review process and UX in github is alien to me. So bear with me if you see unrelated comments every now and then

> Can you configure your editor to not save trailing whitespace?

Sure, I'll push it as a stand alone commit


> So what is your intuition on it? What is the common abstraction between the two?

There are only five call to functions from `ghcup-optparse` component which are all defined in module `GHCup.OptParse.Common`.

- `logGHCPostRm` used in `GHCup.Brick.Actions.del'` function. (I'd link it but I don't know how github works... and it is driving me crazy)
- `fromVersion` used in `GHCup.Brick.Actions.compileHLS` function. 
- `overWriteVersionParser` used as a validator in `CompileGHCMenu`. 
- `ghcVersionTagEither` used as a validator in `CompileHLSMenu`. 

With respect the two lastest bullet points. I think the common (and desirable) abstraction is to factor out validators for arguments.  In general there is a lot of code duplication between workflows via CLI and TUI. Let me explain with GHC compilation workflow:

- In CLI you have `data GHCCompileOptions` and in TUI you have `data CompileGHCOptions` (Yes.. I mess with the names). Both records look almost the same except for the fact that in CLI you have to pass the tool version, and in the TUI you pick it from the state. I am ok having two different, but almost identical, records
- In the CLI, many arguments are parse via functions `String -> Either String a`, and then `optparse-applicative`'s `eitherReader`. In the TUI validators have shape `Text -> Either Text a`. This allow us to give instant feedback in the TUI. I think this part is important to factor out in a common module/component, that way only arguments accepted by the CLI are accepted by the TUI. For example: if CLI expect absolute path, then TUI should do the same.
- The functions compiling GHC are almost identical in the two interfaces. I am ok duplicating this code, because the main differences are in logging, exitcode and monad instances which isn't worth to abstract away. Also, both interfaces rely on `lib:ghcup.GHCup.GHC.compileGHC` function to actually do the work, so the most important part is already abstracted, and only changes in such function will propagate to CLI and TUI.

With respect the first two points... well I don't know how wrothy is abstraction. `logGHCPostRm` was there from the previous version, and `fromVersion` is just part of the `compileHLS` function.

lsmor:
Hi again!

This is pretty much ready to test. The only problem is that I don't know how to test it. I have being able to set up a docker container with both ghcup-v0.1.22.0 and ghcup-this.branch, which set of parameter do I need to test? specially for the compile menus. 

For example, `ghcup compile ghc` has two flags `--make` and `--hadrian` and also `--flavour`. What do they mean? which  are the patches referenced in the flag `--patch`?  etc...

hasufell:
> For example, `ghcup compile ghc` has two flags `--make` and `--hadrian` and also `--flavour`. What do they mean? which are the patches referenced in the flag `--patch`? etc...

`--make` and `--hadrian` are the two build systems of GHC. Make was later decomissioned. So old GHCs only have make, then there are a couple of releases that have both make and hadrian and then the latest ones only have hadrian. I forgot which versions exactly. The selection logic currently is like this:

https://github.com/haskell/ghcup-hs/blob/7daf19961007fe4d1f0565655e43547a4d81877d/lib/GHCup/GHC.hs#L988-L1005

`--patch` are user supplied patches  and can be specified multiple times (file or http/https url, all in URI format). `--patchdir` is similar, except is just a directory path, see the `--help` text.

`--flavour` is documented here: https://gitlab.haskell.org/ghc/ghc/blob/master/hadrian/doc/flavours.md ...but we should accept any string.

hasufell:
It should be `--docs=none`, see https://gitlab.haskell.org/ghc/ghc/-/blob/master/hadrian/doc/user-settings.md#documentation

But I just realize we only let you define the flavor and pass configure args, not hadrian args.

The idea was that "flavour transformers" should be enough: https://gitlab.haskell.org/ghc/ghc/-/blob/master/hadrian/doc/flavours.md#flavour-transformers

But there appears to be none to disable docs.

But you can apply a patch to disable docs in the build flavor:; https://gist.githubusercontent.com/hasufell/8de8890c6ede3ea190a1d177a6cae5ac/raw/e25cf46563bcb37e7b3cb696c25a034d9ebd064a/0001-Disable-docs-for-quickest-build-flavor.patch

Then your line becomes:

```sh
ghcup compile ghc -v 9.8.2 -b 9.4.8 --hadrian --flavour quickest --isolate /root/.cache/host-machine/builds/test -j 4 --patch=https://gist.githubusercontent.com/hasufell/8de8890c6ede3ea190a1d177a6cae5ac/raw/e25cf46563bcb37e7b3cb696c25a034d9ebd064a/0001-Disable-docs-for-quickest-build-flavor.patch
```

hasufell:
Related: https://github.com/haskell/ghcup-hs/issues/846

lsmor:
Hi there,

Good news!. I've been able to compile both `ghc` and `hls` from the TUI. The configurations I've tested are (CLI equiv.)

```
ghcup compile ghc -v 9.8.2 -b 9.4.8 --hadrian --flavour quickest --isolate /root/.cache/host-machine/builds/test -j 4 --patch=https://gist.githubusercontent.com/hasufell/8de8890c6ede3ea190a1d177a6cae5ac/raw/e25cf46563bcb37e7b3cb696c25a034d9ebd064a/0001-Disable-docs-for-quickest-build-flavor.patch

ghcup compile hls -v 2.7.0.0 --isolate /root/.cache/host-machine/builds/test --ghc 9.4.8 -j 4 --cabal-update
```

The combinations of different settings is large. Is there any particular setting I should test?. 

I have created a precarious ubuntu-based dockerfile in [this](https://github.com/lsmor/ghcup-hs/tree/advance-options-docker-test) branch in my fork. The dockerfile should be enough to compile all three of `ghcup`, `ghc` (with no docs) and `hls`. The isolated paths are pointing to mounted volumes in the docker container, so you need to set them up.

hasufell:
> What else is needed to move this PR forward?

Not much. I just need to find the time to go through it again. I'm still recovering from health issues.

lsmor:
> I just need to find the time to go through it again

Let me know if I can facilitate something: documentation, testing workflows, etc...

dfordivam:
I have tried this PR and made some minor fixes on [this branch](https://github.com/dfordivam/ghcup-hs/tree/pr949).

There are a couple of issues I noticed
* The ghcup cannot be compiled without tui. (by disabling `tui` flag in cabal)
* The HLS compilation via ContextMenu is not working for me. I will debug further, and pushes fixes to [this branch](https://github.com/dfordivam/ghcup-hs/tree/pr949).

This is an initial feedback, more testing is necessary.

hasufell:
@lsmor if you happen to be on Matrix, consider joining #ghcup:matrix.org

lsmor:
Thanks for the feedback and for the contibutions!

> * The ghcup cannot be compiled without tui. (by disabling `tui` flag in cabal)

Indeed. I notice the cabal file is too optimistic with respect the tui. Here is the diff you need to make it `tui` independant

```patch
❯ git diff ghcup.cabal
diff --git a/ghcup.cabal b/ghcup.cabal
index 0a9545f..0ed5275 100644
--- a/ghcup.cabal
+++ b/ghcup.cabal
@@ -402,19 +402,13 @@ executable ghcup
   build-depends:
     , ghcup
     , ghcup-optparse
-    , ghcup-tui
 
   if flag(internal-downloader)
     cpp-options: -DINTERNAL_DOWNLOADER
 
   if flag(tui)
     cpp-options:   -DBRICK
-    other-modules: BrickMain
-    build-depends:
-      , brick         ^>=2.1
-      , transformers  ^>=0.5
-      , vty           ^>=6.0 || ^>=6.1 || ^>=6.2
-      , optics        ^>=0.4
+    build-depends: ghcup-tui
 
   if os(windows)
     cpp-options: -DIS_WINDOWS
```

> * The HLS compilation via ContextMenu is not working for me

Which configuration doesn't work for you?


> if you happen to be on Matrix, consider joining #ghcup:matrix.org

I am not in matrix nor any messaging platform. But I could join if there is a discussion about this PR there.

hasufell:
> I am not in matrix nor any messaging platform. But I could join if there is a discussion about this PR there.

Up to you. I'm just trying to gather contributors and establish a less formal place to get pinged, do discussions and ask questions.

E.g. I don't know if you will be at Zurihac.

dfordivam:
@lsmor I did more testing and had the following feedback, plus responses to your comments.

1. In HLS advanced compilation
   * Allow specifying comma or space separated list of multiple target GHC's in the same field. (A single invocation of `ghcup compile ghc` supports specifying multiple GHC values.)
   * "cabal project" / "cabal project local" seems to be not that useful and could be removed (??)
   * a new field for "-g" option would be very useful. ("The git commit/branch/ref to build from (accepts anything 'git checkout' accepts)"

   
2. Print the CLI command which would do the exact advanced installation as being performed by the tui

    If the user is trying to do something experimental the possibility of the compilation failing is high and the user will have to check the logs and fire the compilation again.
    in this case providing the user with the CLI command with all the fields specified would be nice.

    This could be printed as an "Info"



3. > Which configuration doesn't work for you?

    I hit the following error when building HLS, and interestingly I cannot reproduce this now.

```
[ Info  ] verifying digest of: haskell-language-server-2.7.0.0-src.tar.gz
[ Info  ] Unpacking: haskell-language-server-2.7.0.0-src.tar.gz to /home/divam/vms/ubuntu/.ghcup/tmp/ghcup-4f98de77d2a77843
[ Info  ] Installing HLS
ghcup: user error (Pattern match failure in 'do' block at lib/GHCup/HLS.hs:231:3-12)
```

4. In response to [this comment](https://github.com/haskell/ghcup-hs/commit/a4d278786b04e5e084d287ee0cb8ad19919dd6db#commitcomment-142406113)

Both of the APIs return `Either String ()`, which is handled in `withIOAction`

If this API returns a `Right` value in case of an error, then "Success" is printed like this

```
[ Error ] [GHCup-00841] Process "/home/divam/vms/ubuntu/.ghcup/tmp/ghcup-b01477984640547e/ghc-9.10.1/hadrian/build" with arguments ["-j8",
[ ...   ]                                                                                                             "binary-dist"] failed with exit code 1.
Success
Press enter to continue
```

Whereas if a `Left` value is returned the following is printed

```
[ ghc-conf ] configure: error: GHC version 9.6 or later is required to compile GHC.
Error: [GHCup-00841] Process "sh" with arguments ["./configure", "GHC=ghc-9.8.2",
                             "--prefix=/home/divam/vms/ubuntu/.ghcup/ghc/9.10.0.20240328",
                             "--disable-ld-override"] failed with exit code 1.
Press enter to continue
```

5. In response to [this comment](https://github.com/haskell/ghcup-hs/commit/bb4c9a32d8f8410f9784d74514d3513e3f57e111#commitcomment-142406086)

The most likely use cases for the tui based advanced installations would be "simple" scenarios like compiling HLS (when a new minor release of GHC is released), and the power users would prefer the CLI.

If the user is invoking the advanced options menu for compile/install of only a single version of a tool, then keeping the state is certainly a better choice, since the user might want to tweak the values in case the installation or compilation did not happen successfully.

(Note: User should be able to run the compilation of HLS for multiple GHC versions in a single step, as I mentioned above in this comment)


When the user want to use the advanced options submenu multiple times in a
single session of GHCup, then the design choice for keeping/resetting the state becomes more complicated.

Therefore let's consider both of these design choices for a couple of scenarios I think would be common.

* Install multiple versions of GHC and specify an isolated directory for each of them.

   After the installation finishes, user will open the advanced install menu for another version of GHC, and will have to specify the isolated directory again.

   In this scenario I think both keeping the state / clearing the state should be fine, since user has to specify only a single field.
   (Keeping the state has an advantage that the user might want to just change the last part of the path, edit it manually, and fire the install again.)


* Compile multiple version of GHC, with CONFIGURE_ARGS or flavour specified

   For example, user wants to build a static flavor for multiple versions.
   In this case the user will likely specify these fields: bootstrap-ghc, overwrite-version, jobs, flavour, and isolated.

   For compiling different GHC the bootstrap-ghc and isolated fields, will be different.
   But having the flavor and jobs fields preserved is good.
   For overwrite-version field, it depends whether its a simple string or more complex like '%v-%h'
   Therefore both keeping the state and reseting has pros/cons.

I think other than these two scenarios there aren't many use cases for using the advanced options menu. Anything more complicated might be better done via CLI.

Overall I think maintaining the state is better.

lsmor:
> * Allow specifying comma or space separated list of multiple target GHC's in the same field.

Oh!! sure, I'll do it

> * "cabal project" / "cabal project local" seems to be not that useful and could be removed (??)

I did not commit those... 

> * a new field for "-g" option would be very useful.

I remmember to implement this, but there was a problem with it... not very sure now, I'll take a look a come back with feedback later.

I will merge your commits and push them (read below). By the way,  thanks a lot for the help and testing!!! probably keeping the state is better than not keeping it. 

> At any rate... I believe the TUI can very well be improved gradually

Absolutely, I was thinking about adding a section to `docs/dev.md` with the docs for the TUI, so hopefully (unrealisticly?) other people can contribute.

Now a git question... how do I proceed? can @dfordivam merge to this PR directly? o should I add locally his fork to my remotes, then pull, then push to my fork? (the last is moreless what I do at job, but the usage of git is much more amateur)


To recap: 

- [ ] add multiple version of ghc for `hls` compilation
- [ ] explore the -g option and implement it if is easy enough.
- [ ] merge @dfordivam job on this.

dfordivam:
> I did not commit those...

I was only referring the TUI interface. The TUI could provide a different set of options than CLI.
But since these options have already been added in the TUI, it's fine to keep them.
Although I think they will be hardly ever used, they could be moved to the bottom, which I did on my branch.
 
> should I add locally his fork to my remotes, then pull, then push to my fork..

This is how it's usually accomplished, but I opened [a PR on your fork](https://github.com/lsmor/ghcup-hs/pull/1) to make it easy for you to merge the branch via GitHub.

I have also pushed a few more commits on my branch, including the cabal file fixes.
In case there are merge conflict with your local changes, or you don't want to merge all the changes, you should fetch my branch and cherry-pick commits on top of your branch, and then push to this PR.

lsmor:
Hi there:

I have rebased master and think about @dfordivam feedback. In particular, these two suggestions:

> * Allow specifying comma or space separated list of multiple target GHC
> * a new field for "-g" option would be very useful.

The current design is that the target to compile is the version the cursor is over. So if you open the context menu of `ghc-9.8.2` you'd compile that precise version. IMO, It would be a little bit weird if you open the menu but you can compile whatever you want. The same argument applies to cli's `-g`  option, which is disable in both `hls` and `ghc` context menus.

I think is ok if this PR includes most of the functionality of the cli provides, sacrificing it a little bit of it for a better UX.

> I don't know if you will be at Zurihac

I wasn't... sorry for the late response. :sweat_smile:

dfordivam:
Hi @lsmor 

> The current design is that the target to compile is the version the cursor is over. So if you open the context menu of ghc-9.8.2 you'd compile that precise version.

I think there is a slight confusion here, if you put the cursor on GHC, the advanced compile menu will open the options for compiling the GHC.

The user needs to put the cursor over HLS to open the options for compiling HLS.
In this case the target GHC will not be chosen, and the user will have to manually fill in the target GHC version(s).

_"Allow specifying comma or space separated list of multiple target GHC"_ option is only valid in the case of HLS compilation, as HLS can be compiled for multiple GHC targets. For GHC this does not make sense.


Nevertheless, this is a usability related issue, and both of these can be tackled later, after merging this PR.
It may be nice to even create another advanced option menu, which allows the user to compile HLS when the cursor is over GHC, but that can also be done later.

The remaining fixes necessary for this PR are:


- There are a few hlint warnings, should be straightforward to fix.

- Tests are failing to compile, this should also be straightforward to fix

```
 test/optparse-test/SetTest.hs:27:13: error:
    • Data constructor not in scope:
        SetRecommended :: GHCup.Utils.Parsers.SetToolVersion
    • Perhaps you meant ‘Recommended’ (imported from GHCup.Types)
   |
27 |   [ ("set", SetRecommended)
   |             ^^^^^^^^^^^^^^
```

lsmor:
> The remaining fixes necessary for this PR are:
> 
> * There are a few hlint warnings, should be straightforward to fix.
> * Tests are failing to compile, this should also be straightforward to fix

Done. There is a check failing but is I think is a problem with the host machine:

```
[ 1 of 10] Compiling Data.Hashable.FFI ( src/Data/Hashable/FFI.hs, dist/build/Data/Hashable/FFI.o, dist/build/Data/Hashable/FFI.dyn_o )
clang-13: error: the clang compiler does not support '-march=native'

<no location info>: error:
    `clang' failed in phase `C Compiler'. (Exit code: 1)
```

Up to internet, is because of Apple's M1 chip

ghcup.cabal
```diff
@@ -354,9 +406,8 @@ executable ghcup
     other-modules: BrickMain
```
dfordivam, line 406: once you remove the app/ghcup/BrickMain.hs, Remove this line also

app/ghcup/Main.hs
```diff
@@ -11,7 +11,8 @@
 module Main where
 
 #if defined(BRICK)
-import           BrickMain                    ( brickMain )
+-- import           BrickMain                    ( brickMain )
```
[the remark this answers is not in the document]
lsmor: done!

ghcup.cabal
lsmor: Yes, also, the executable can drop pedencies on vty/brick because now those are in the ghcup-tui library.

dfordivam (approved):

hasufell:
Thanks @lsmor. Great work.

Issue #952, opened 2023-12-15
Windows installation fails during "Downloading Msys2 archive"-step (`WriteErrorException`)


runeksvendsen:
I just installed a fresh copy of Windows 11 Pro in a VM, and ran the command from https://www.haskell.org/ghcup/# that appears under "To install on Windows" after clicking "Show all platforms".

Relevant error:

```
Extracting Msys2 archive...
Set-ExecutionPolicy Bypass -Scope Process -Force;[System.Net.ServicePointManager]::SecurityProtocol =
[System.Net.ServicePointManager]::SecurityProtocol -bor 3072; try { Invoke-Command -ScriptBlock
([ScriptBlock]::Create((Invoke-WebRequest https://www.haskell.org/ghcup/sh/bootstrap-haskell.ps1
-UseBasicParsing))) -ArgumentList $true } catch { Write-Error $_ } : Cannot remove item
C:\Users\rune\AppData\Local\Temp\msys2-base-x86_64-20221216.sfx.exe: Access to the path
'C:\Users\rune\AppData\Local\Temp\msys2-base-x86_64-20221216.sfx.exe' is denied.
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException
```

**EDIT:** *Looks like this is the line that causes the failure:* https://github.com/haskell/ghcup-hs/blob/d05fad49a17d82523656d64a2ec82af3c3c7fc3f/scripts/bootstrap/bootstrap-haskell.ps1#L416

Full logs:

```
PS C:\Users\rune> Set-ExecutionPolicy Bypass -Scope Process -Force;[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; try { Invoke-Command -ScriptBlock ([ScriptBlock]::Create((Invoke-WebRequest https://www.haskell.org/ghcup/sh/bootstrap-haskell.ps1 -UseBasicParsing))) -ArgumentList $true } catch { Write-Error $_ }
Picked C:\ as default Install prefix!
Welcome to Haskell!

This script can download and install the following programs:
  * ghcup - The Haskell toolchain installer
  * ghc   - The Glasgow Haskell Compiler
  * msys2 - A linux-style toolchain environment required for many operations
  * cabal - The Cabal build tool for managing Haskell software
[...]

PR #954, opened 2023-12-20
bootstrap-haskell.ps1: don't exit if tmp file removal fails

runeksvendsen:
Fixes #952

PR #956, opened 2023-12-23
bootstrap-haskell.ps1: don't exit if tmp file removal fails

runeksvendsen:
Fixes https://github.com/haskell/ghcup-hs/issues/952

Replaces #954


hasufell (approved):

hasufell:
deployed

Issue #962, opened 2024-01-02
ghcup binary built on alpine-3.16 with GHC-9.4.8 segfaults


hasufell:
The bindist used is: https://downloads.haskell.org/~ghcup/unofficial-bindists/ghc/9.4.8/ghc-9.4.8-i386-alpine-linux.tar.xz

It is a dynamically linked bindist, because the static ones are more broken.

<details>
  <summary>The settings file is as follows</summary>
  
```
[("GCC extra via C opts", "")
,("C compiler command", "gcc")
,("C compiler flags", "-U__i686")
,("C++ compiler command", "g++")
,("C++ compiler flags", "")
,("C compiler link flags", "")
,("C compiler supports -no-pie", "YES")
,("Haskell CPP command", "gcc")
,("Haskell CPP flags", "-E -undef -traditional")
,("ld command", "ld")
,("ld flags", "")
,("ld supports compact unwind", "NO")
,("ld supports build-id", "YES")
,("ld supports filelist", "NO")
,("ld is GNU ld", "YES")
,("Merge objects command", "ld")
,("Merge objects flags", "-r")
,("ar command", "ar")
,("ar flags", "q")
,("ar supports at file", "YES")
,("ar supports -L", "NO")
,("ranlib command", "ranlib")
,("otool command", "otool")
,("install_name_tool command", "install_name_tool")
,("touch command", "touch")
,("dllwrap command", "/bin/false")
,("windres command", "/bin/false")
,("libtool command", "libtool")
,("unlit command", "$topdir/bin/unlit")
,("cross compiling", "NO")
,("target platform string", "i386-unknown-linux")
,("target os", "OSLinux")
,("target arch", "ArchX86")
,("target word size", "4")
,("target word big endian", "NO")
,("target has GNU nonexec stack", "YES")
,("target has .ident directive", "YES")
,("target has subsections via symbols", "NO")
,("target has RTS linker", "YES")
,("target has libm", "YES")
,("Unregisterised", "NO")
,("LLVM target", "i686-unknown-linux")
,("LLVM llc command", "llc")
,("LLVM opt command", "opt")
,("LLVM clang command", "clang")
,("Use inplace MinGW toolchain", "NO")
,("Use interpreter", "YES")
,("Support SMP", "YES")
,("RTS ways", "debug thr thr_debug thr_p dyn debug_dyn thr_dyn thr_debug_dyn thr_debug_p debug_p")
,("Tables next to code", "YES")
[...]

PR #968, opened 2024-01-04
Set LD=ld.bfd on Alpine linux during bindist configure

### LINKED ISSUE #967: Default to `ld.bfd` for bindist configure (alpine or all linux?)
GHC likes to pick `ld.gold`. However, experience shows that it's more buggy and unreliable, e.g. https://gitlab.haskell.org/ghc/ghc/-/issues/17508

GHCup strives to give users a configuration that works reliably, not optimize for compilation speed and risk segfaults.

The idea would be to do the following on Alpine or even all Linuxes:

* check if there's a `ld.bfd` binary present
* check if `LD` environment variable is not set
* if both is true, pass `LD=ld.bfd` to the environment executing the bindist configure script

hasufell:
Fixes #967

### DIFF
--- a/lib/GHCup/GHC.hs
+++ b/lib/GHCup/GHC.hs
@@ -449,13 +449,30 @@ installUnpackedGHC path inst tver forceInstall addConfArgs
            = []
 
       lift $ logInfo "Installing GHC (this may take a while)"
+      env <- case _rPlatform of
+               -- https://github.com/haskell/ghcup-hs/issues/967
+               Linux Alpine
+                 -- lets not touch LD for cross targets
+                 | Nothing <- _tvTarget tver -> do
+                     cEnv <- liftIO getEnvironment
+                     spaths <- liftIO getSearchPath
+                     has_ld_bfd <- isJust <$> liftIO (searchPath spaths "ld.bfd")
+                     let ldSet = isJust $ lookup "LD" cEnv
+                     -- only set LD if ld.bfd exists in PATH and LD is not set
+                     -- already
+                     if has_ld_bfd && not ldSet
+                     then do
+                       lift $ logInfo "Detected alpine linux... setting LD=ld.bfd"
+                       pure $ Just (("LD", "ld.bfd") : cEnv)
+                     else pure Nothing
+               _ -> pure Nothing
       lEM $ execLogged "sh"
                        ("./configure" : ("--prefix=" <> fromInstallDir inst)
                         : (maybe mempty (\x -> ["--target=" <> T.unpack x]) (_tvTarget tver) <> ldOverride <> (T.unpack <$> addConfArgs))
                        )
                        (Just $ fromGHCupPath path)
                        "ghc-configure"
-                       Nothing
+                       env
       tmpInstallDest <- lift withGHCupTmpDir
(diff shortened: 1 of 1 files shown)


runeksvendsen (commented):
How much work would it be to test in CI that this fixes the segfault issue?

runeksvendsen:
This also fixes https://github.com/haskell/ghcup-hs/issues/962, right?

hasufell:
> This also fixes #962, right?

No, it unfortunately doesn't. I'm suspecting an actual bug in 9.4.8 series.

hasufell:
> How much work would it be to test in CI that this fixes the segfault issue?

I don't feel that's gonna add much.

We have evidence here:

* https://gitlab.haskell.org/ghc/ghc/-/issues/17508#note_313450
* https://sourceware.org/bugzilla/show_bug.cgi?id=23856

The minimal reproducer (C code) is shown here (and still works): https://gitlab.haskell.org/ghc/ghc/-/issues/17508#note_372662

There's no minimal reproducer for Haskell.

My past days as Gentoo distributor also remind me that ld.gold was always more problematic, so I don't think it's a good default to begin with.

hasufell:
> There's no minimal reproducer for Haskell.

Ah, silly me... I didn't edit the settings file correctly.

Obviously, this works:

```
$ cat Foo.hs
main = putStrLn "hello world"
ghc Foo.hs
./Foo
Segmentation fault (core dumped)
```

Then make sure you ghc settings file has these lines:

```
,("C compiler link flags", "-fuse-ld=gold")
,("ld command", "ld.gold")
,("Merge objects command", "ld.gold")
```

runeksvendsen (commented):
The way I read the code, we previously ran the configure script with an empty environment — ie. it did not inherit anything from the environment of GHCup. Is this correctly understood?

If this is the case then I'm not convinced that we should pass all of GHCup's environment plus `LD=ld.bfd` to the configure script in the positive case. In that case why not use an environment with only `LD=ld.bfd`?

runeksvendsen:
> > we previously ran the configure script with an empty environment — ie. it did not inherit anything from the environment of GHCup. Is this correctly understood?
> 
> No, that's not correct. That parameter overwrites the current environment for the subprocess. `LD=ld.bfd ghcup install ghc ...` works perfectly fine.

So passing `Just [("LD", "ld.bfd")]` as the last argument to `execLogged` will execute the process in the same environment as that of GHCup but with the LD env var overwritten?

If that's the case then why do we need

```
(("LD", "ld.bfd") : cEnv)
```

instead of just

```
[("LD", "ld.bfd")]
```

?

hasufell:
> So passing Just [("LD", "ld.bfd")] as the last argument to execLogged will execute the process in the same environment as that of GHCup but with the LD env var overwritten?

No, you pass the entire environment.

runeksvendsen:
Okay I understand now. I thought passing `Nothing` as the environment means the executed process would have an empty environment. But `Nothing` means "inherit environment of the parent process" whereas `Just []` means "use an empty environment".

runeksvendsen (approved):
The code looks good to me. But I don't have a way to test it, so I will leave that up to you.

PR #976, opened 2024-01-13
Use STACK_XDG env to determine the stack root path

gxmas:
During installation, a hook script is copied to the hooks directory under _stack_'s root path.  The aim is to provide a better integration between _ghcup_ and _stack_.  

Currently, the installation script only considers the `STACK_ROOT` env variable to determine _stack_'s root path.  _Stack_, however, can also be configured to follow the XDG base directory specification by setting the `STACK_XDG` env, as explained [here](https://docs.haskellstack.org/en/stable/stack_root/).

For people using _stack_ with the XDG base directory specification, the hook script ends up being copied in the default _stack_'s root path, i.e. in `${HOME}/.stack`, resulting in a lack of integration between _ghcup_ and _stack_.

This change takes into account the `STACK_XDG` env to determine the proper _stack_ root path.

### DIFF
--- a/scripts/bootstrap/bootstrap-haskell
+++ b/scripts/bootstrap/bootstrap-haskell
@@ -738,6 +738,15 @@ ask_stack() {
 	unset stack_answer
 }
 
+find_stack_root() {
+    if [ -n "${STACK_ROOT}" ] ; then
+        echo "${STACK_ROOT}"
+    elif [ -n "${STACK_XDG}" ] ; then
+        echo "${XDG_DATA_HOME}/stack"
+    else
+        echo "${HOME}/.stack"
+    fi
+}
 
 find_shell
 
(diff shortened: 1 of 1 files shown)


scripts/bootstrap/bootstrap-haskell
```diff
@@ -738,6 +738,15 @@ ask_stack() {
 	unset stack_answer
 }
 
+find_stack_root() {
+    if [ -n "${STACK_ROOT}" ] ; then
+        echo "${STACK_ROOT}"
+    elif [ -n "${STACK_XDG}" ] ; then
+        echo "${XDG_DATA_HOME}/stack"
```
hasufell, line 745: `XDG_DATA_HOME` is not always set.

The XDG specification says to assume `$HOME/.local/share` if it is not set.

gxmas: Added `$HOME/.local/share` as a default value for `$XGD_DATA_HOME` if not set.

hasufell:
This has been deployed.

PR #977, opened 2024-01-16
Disable cabal XDG by default

hasufell:
https://github.com/haskell-actions/setup/issues/62


runeksvendsen:
Relevant comment: https://github.com/haskell-actions/setup/issues/62#issuecomment-1894035317

PR #987, opened 2024-01-22
Restore TUI scrolling to old behavior

hasufell:
@lsmor 

As discussed: https://github.com/haskell/ghcup-hs/pull/850#issuecomment-1903553150

### DIFF
--- a/app/ghcup/BrickMain.hs
+++ b/app/ghcup/BrickMain.hs
@@ -11,6 +11,7 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE TemplateHaskell   #-}
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE ViewPatterns          #-}
 
 module BrickMain where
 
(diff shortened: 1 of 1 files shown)


lsmor:
I think this does not recovers completelly old behaviour since it jumps if the section below doesn't fit on the screen

hasufell:
> I think this does not recovers completelly old behaviour since it jumps if the section below doesn't fit on the screen

It doesn't jump here:

Peek 2024-01-22 22-42

app/ghcup/BrickMain.hs
```diff
@@ -202,41 +204,42 @@ handleGenericListEvent (VtyEvent ev) = do
...
-                                in  (acc_widget <=> new_widget, acc_off + new_off)
-                                )
-                (Brick.emptyWidget, 0)
-                elms
-        Brick.render $  Brick.viewport sl_name Brick.Vertical $ Brick.translateBy (Brick.Location (0, min 0 (limit-off))) widget
+renderSectionList renderElem sectionFocus ge@(GenericSectionList focus elms slName) =
+    Brick.Widget Brick.Greedy Brick.Greedy $ Brick.render $ Brick.viewport slName Brick.Vertical $
+      V.ifoldl' (\(!accWidget) !i list ->
+                      let hasFocusList = sectionIsFocused list
+                          makeVisible = if hasFocusList then Brick.visibleRegion (Brick.Location (c, r)) (1, 1) else id
```
hasufell, line 216: Here's the trick: we compute `r` (the row-location relative to the current element) based on the focused element of the focused section (see below)

```hs
  -- compute the location to focus on within the active section
  (c, r) :: (Int, Int) = case sectionListSelectedElement ge of
                           Nothing -> (0, 0)
                           Just (selElIx, _) -> (0, selElIx)
```

And then use the smallest possible dimension to focus on `(1, 1)`.

lsmor:
> It doesn't jump here:

Probably somthing weird with the vscode terminal. It doesn't jump on gnome's. In case you wonder It was a little bit easier to do, since in the version at #850 the row-location was already calculated in the variable `off`. By using the trick of minimizing the display area `(1,1)` you can use `visibleRegion` to push a request visibility for the current row

```haskell
renderSectionList render_elem section_focus (GenericSectionList focus elms sl_name) =
    Brick.Widget Brick.Greedy Brick.Greedy $ do
        c <- Brick.getContext
        let -- A section is focused if the whole thing is focused, and the inner list has focus
            section_is_focused l = section_focus && (Just (L.listName l) == F.focusGetCurrent focus)
            s_idx = fromMaybe 0 $ V.findIndex section_is_focused elms
            render_inner_list has_focus l = Brick.vLimit (length l) $ L.renderList (\b -> render_elem (b && has_focus)) has_focus l
            (widget, off) = 
                V.ifoldl' (\wacc i list ->
                                let has_focus_list = section_is_focused list
                                    (!acc_widget, !acc_off) = wacc
                                    new_widget = if i == 0 
                                                 then render_inner_list has_focus_list list
                                                 else hBorder <=> render_inner_list has_focus_list list
                                    new_off 
                                        | i < s_idx  = 1 + L.listItemHeight list * length list 
                                        | i == s_idx = 1 + L.listItemHeight list * fromMaybe 0 (L.listSelected list)
                                        | otherwise  = 0
                                in  (acc_widget <=> new_widget, acc_off + new_off)
                                )
                (Brick.emptyWidget, 0)
                elms
        Brick.render 
          $ Brick.viewport sl_name Brick.Vertical
          -- This is the only change needed. for some reason off isn't enoughm, but off-1 does the job
          $ Brick.visibleRegion (Brick.Location (0, off - 1)) (1,1) widget
```

hasufell:
> the row-location was already calculated in the variable off

That's the row-location across all sections afaiu, not just the currently active section. I think we don't need that calculation anymore, since we're not translating across all sections.

hasufell:
@lsmor can you accept the  repo acces invite so I can add you as reviewer?

app/ghcup/BrickMain.hs
lsmor: Honestly, I'm quite puzzled that this works... hahaha afaiu `(c,r)` are relative to the inner section, so when we call `Brick.visibleRegion (Location (c, r)) ...` it sends a visibility request to the outter viewport (a viewport for the whole section-list) using the coordinates of the inner section... Any case, It works and it is very clear

lsmor (approved):

runeksvendsen (approved):
I'm not convinced it would make sense for me to spend time reading and understanding the code, so I just compiled and ran the TUI locally, and can confirm that it fixes the issue in question.

hasufell:
I dropped the whitespace commit

Issue #1008, opened 2024-02-19
`stack.yml.lock` and BUILD.md to tell user how to build ghcup from scratch


evan0greenup:
It would be nice to provide a `stack.yml.lock` and a `BUILD.md` markdown file to teach users who to build `ghcup-hs` from scratch.

hasufell:
Does `cabal build` or `stack build` not work?

evan0greenup:
@hasufell , Both not working, the reason is due to the version of dependencies.

hasufell:
Yeah, see https://github.com/haskell/ghcup-hs/pull/1007

stevenleeS0ht:
@evan0greenup , @hasufell , It seems that the reason why `stack build` fail is due to the dependency list in `stack.yml` is outdate. I create a pull request. https://github.com/haskell/ghcup-hs/pull/1009

Please have a review and approve it.

hasufell:
This is fixed now.

PR #1009, opened 2024-02-19
dependency list bugfix in stack.yml

stevenleeS0ht:
To resolve the issue of this https://github.com/haskell/ghcup-hs/issues/1008


hasufell:
This is already in progress in https://github.com/haskell/ghcup-hs/pull/1007 as explained in the ticket

PR #1027, opened 2024-03-14
Update stack.yaml to resolve `stack build` failure

stevenleeS0ht:
Resolve this #1026 .

PR #1029, opened 2024-03-16
Fix Stack CI

hasufell:


hasufell:
Supersedes https://github.com/haskell/ghcup-hs/pull/1027

PR #1052, opened 2024-04-29
Migrate back to conduit, fixes #1051

### LINKED ISSUE #1051: Move away from streamly?
Streamly has been rather difficult to maintain:

- it breaks API often, requiring major updates/rewrites
   - https://github.com/hasufell/streamly-yaml/pull/5
- it makes improper use of cabal flags, which has been communicated several times to the maintainers
   - https://github.com/haskell/cabal/issues/8128
- sometimes older releases are incompatible with new GHC versions and the compatiblity patch is mushed into the next major API version, giving no room for minimal updates
   - https://github.com/composewell/streamly/pull/2415#issuecomment-1913069210
- currently doesn't build for us at all with GHC-9.6.5:
   - https://github.com/composewell/streamly/issues/2738
   - https://github.com/composewell/streamly/issues/2739
- I have to maintain streamly variants of libyaml to avoid unnecessarily depending on e.g. conduit:
   - https://hackage.haskell.org/package/yaml-streamly

All this is maintenance cost. At the moment I don't see the advantage of using streamly anymore. I'm thinking to move to another streaming framework, maybe back to conduit.

hasufell:


hasufell:
https://hackage.haskell.org/package/conduit-1.3.5/docs/src/Data.Conduit.Combinators.html#sourceDirectoryDeep is trash: it calls `getFileType` on every file. Our current implementation avoids doing additional syscalls.

PR #1053, opened 2024-04-29
Use streamly-core instead of streamly

adithyaov:


adithyaov:
@hasufell Can you please approve the workflow?
I'm able to build and test this locally.

adithyaov:
Ah, did not see https://github.com/haskell/ghcup-hs/pull/1052
@hasufell Feel free to close this if you do ahead with reverting back to conduit.

hasufell:
Thanks for the support, but I decided to move to conduit, since my use of streams is very narrow and building across many GHC versions is a higher priority for this project.

PR #1054, opened 2024-04-30
Attempt to fix darwin release

mpickering:
The build works locally for me on a darwin machine (running the `.github/scripts/build.sh` script). I am unable to test this myself on CI it seems.


hasufell:
This doesn't explain the regression between GHC 9.4.x and 9.6.x suddenly unable to use brews toolchain. I'm certain there will be users who do have brew llvm in their path. Why did it stop working?

PR #1062, opened 2024-05-07
Fix the bootstrap script for armv7*

dfordivam:
The current bootstrap script fails with this error. I have only tested this PR on armv7l, and haven't tested it on windows (yet).

```
$ curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
Unknown architecture: armv7l
```


dfordivam:
This PR is just to trigger checks on CI, while I test it manually on windows.

dfordivam:
I tested manually on windows, and this works fine. (and just realized that the CI has windows testing for bootstrap script already)

hasufell (approved):

PR #1072, opened 2024-05-31
Add docs for supporting multiple user install on windows

### LINKED ISSUE #1070: [DOC] Install globally
This comes up frequently enough that we should add it to FAQ, both for linux and windows: https://discourse.haskell.org/t/install-haskell-development-environment-fo-all-users-in-windows/7520/9

Maybe under "Tips and Tricks"? https://www.haskell.org/ghcup/guide/#tips-and-tricks

dfordivam:
I tried these instructions by creating multiple user accounts.
It seems that settings the `PATH` is sufficient for the standard `C:\ghcup` installations.

For non-standard installation `GHCUP_INSTALL_BASE_PREFIX` is necessary, to ensure that the ghcup exe does not create / write to `C:\ghcup`

I am not sure how the `GHCUP_MSYS2` + cabal related instructions are to be tested.

Fixes #1070

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -766,6 +766,44 @@ ghcup run --ghc 8.10.7 --cabal latest --hls latest --stack latest --install -- c
 
 This will execute vscode with GHC set to 8.10.7 and all other tools to their latest version.
 
+## Support multiple users on Windows
+
+On Windows, by default GHCup is installed in `C:\ghcup`, and not in the `C:\Users` directory.
+
+To allow all users of a Windows machine to make use of GHCup all you have to do is set the right global environment variables (and possibly file permissions).
+
+```sh
+$env:PATH += ";C:\ghcup\bin"
+```
+
+If the GHCup is not installed at the `C:\ghcup`, then additionally the following environment variables would also need adjustment
+
+* GHCUP_INSTALL_BASE_PREFIX (to e.g. D:\)
+* GHCUP_MSYS2 (to e.g. D:\ghcup\msys64)
+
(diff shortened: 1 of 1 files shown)


dfordivam:
Ok, but this didn't work smoothly if the ghcup is not installed at 'C:/'.

It is possible to set the env var globally during install
`[Environment]::SetEnvironmentVariable("GHCUP_INSTALL_BASE_PREFIX", 'C:\somedir', [System.EnvironmentVariableTarget]::Machine)`

But if the `GHCUP_INSTALL_BASE_PREFIX` is not defined as user's env variable, then `ghcup` continues to work with `C:\` prefix.

hasufell:
> But if the `GHCUP_INSTALL_BASE_PREFIX` is not defined as user's env variable, then `ghcup` continues to work with `C:\` prefix.

That would appear to be a bug, no?

dfordivam:
Yeah, likely an issue with `lookupEnv`. I looked around but couldn't find any more info about this issue.  Or perhaps `lookupEnv` is fine, and this should be some other windows specific API?

hasufell:
https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-getenvironmentvariablew

> This function can retrieve either a system environment variable or a user environment variable.

dfordivam:
Ok, after more debugging figured out the issue, the powershell needs to be restarted for all users after the install to load the new vars from global.

dfordivam:
I am doing a rewrite of this to be clearer

hasufell:
@dfordivam sorry for the mess... can you re-raise this at https://github.com/haskell/ghcup-www

PR #1077, opened 2024-06-04
Ignore content length checks when the stream is encoded

### LINKED ISSUE #1055: Internal downloader throws `ContentLengthError`
```
$ ghcup --downloader=internal tui
[ Info  ] downloading: https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-0.0.8.yaml as file /home/hasufell/.ghcup/cache/ghcup-0.0.8.yaml
[============================================================================================================================================] 141%
ghcup: ContentLengthError Nothing (Just 98256) 69674
```

dfordivam:
For `tar.xz` files the progress bar + length checks works as expected as they are sent by the server without any encoding.

The yaml files are usually gzipped encoded and the progress bar is not shown.

fixes #1055

### DIFF
--- a/lib/GHCup/Download/IOStreams.hs
+++ b/lib/GHCup/Download/IOStreams.hs
@@ -123,11 +123,16 @@ downloadInternal = go (5 :: Int)
 
     downloadStream r i' = do
       void setup
-      let size = case getHeader r "Content-Length" of
-            Just x' -> case decimal $ decUTF8Safe x' of
+      -- We can only do content length checks & progress bar when the stream is not encoded
+      let notEncoded = case mk <$> getHeader r "Content-Encoding" of
+            Nothing -> True
+            Just "identity" -> True
+            Just _ -> False
+          size = case (notEncoded, getHeader r "Content-Length") of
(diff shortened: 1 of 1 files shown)


lib/GHCup/Download/IOStreams.hs
```diff
@@ -123,11 +123,12 @@ downloadInternal = go (5 :: Int)
 
     downloadStream r i' = do
       void setup
-      let size = case getHeader r "Content-Length" of
-            Just x' -> case decimal $ decUTF8Safe x' of
+      -- We can only do content length checks & progress bar when the stream is not encoded
```
hasufell, line 126: Afaiu  there's also `identity` encoding, which would work.

hasufell: See https://www.iana.org/assignments/http-parameters/http-parameters.xhtml

dfordivam: fixed

dfordivam:
this is ready for review

PR #1087, opened 2024-06-25
Remove `includes` from `.cabal`-file

sol:
This is useless at best, and a common source of confusion.

https://github.com/haskell/cabal/pull/10145
https://github.com/sol/hpack/issues/355


hasufell:
What exactly is the reasoning to remove this?

This field is clearly documented: https://cabal.readthedocs.io/en/stable/cabal-package-description-file.html#pkg-field-includes

PR #1090, opened 2024-06-27
Context menu: add menu validation, git-ref support, minor fixes

dfordivam:
@lsmor Here are some fixes and improvements for context menu, some of these we discussed previously on PR #949 review

It may be easier to review each commit separately.

### DIFF
--- a/lib-tui/GHCup/Brick/Common.hs
+++ b/lib-tui/GHCup/Brick/Common.hs
@@ -47,7 +47,7 @@ module GHCup.Brick.Common  (
     , PatchesEditBox, CrossTargetEditBox, AddConfArgsEditBox, OvewrwiteVerEditBox
     , BuildFlavourEditBox, BuildSystemEditBox, OkButton, AdvanceInstallButton
     , CompileGHCButton, CompileHLSButton, CabalProjectEditBox
-    , CabalProjectLocalEditBox, UpdateCabalCheckBox
+    , CabalProjectLocalEditBox, UpdateCabalCheckBox, GitRefEditBox
   ) ) where
 
 import           GHCup.List ( ListResult )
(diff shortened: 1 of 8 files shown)


lib-tui/GHCup/Brick/Common.hs
lsmor, line 1: Is there any problem If the user selects a GitRef which belongs to a version of the compiler other that the one the user has the cursor over  in the TUI? I guess no, but I was hesatite to include the git-ref when building myself, because of this sort of problem

lib-tui/GHCup/Brick/Widgets/Menu.hs
lsmor, line 1: This is a very good idea, indeed

lib-tui/GHCup/Brick/Common.hs
lsmor, line 1: Is there any problem If the user selects a GitRef which belongs to a version of the compiler other that the one the user has the cursor over  in the TUI? I guess no, but I was hesatite to include the git-ref when building myself, because of this sort of problem

lsmor (approved):

lib-tui/GHCup/Brick/Common.hs
dfordivam: From the cursor position we were only obtaining the version of tool.
 (and whether we want to show the compile HLS/GHC menu)
 
I don't see any problem if the user specifies a different version of the tool,
 and we cannot guarantee that the build will succeed based on the
 options user have provided. So it's up to the user to try it out and see if compilation works for an arbitrary git ref.

dfordivam: I did try this menu with compiling the GHC's wip backports branch, and it worked fine.

lsmor: nice, great improvements!

PR #1091, opened 2024-07-01
Context menu: Make UI compact

dfordivam:
This makes the Compile GHC / HLS menus more compact, so that all the fields can be now displayed without scrolling even on a considerably smaller terminal height.

2024-07-01-155952_screenshot

2024-07-01-155803_screenshot

2024-07-01-155836_screenshot


dfordivam:
The button could be aligned right. Though it looks fine left aligned as well.

2024-07-01-161144_screenshot

Issue #836, opened 2023-06-10
Unable to install on Windows - curl mysys2 error


parkeradam:
Using the command from https://www.haskell.org/ghcup/# for windows, I consistently without fail get the error of the following. 

```
...Msys2 doesn't exist, installing into C:\\ghcup\msys64
Starting installation in 5 seconds, this may take a while...
Downloading Msys2 archive 20221216...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (35) schannel: next InitializeSecurityContext failed: Unknown error (0x80092012) - The revocation function was unable to check revocation for the certificate.
Exception: Exec: Error executing command curl.exe with arguments '-o C:\Users\mrpar\AppData\Local\Temp\\msys2-base-x86_64-20221216.sfx.exe https://repo.msys2.org/distrib/x86_64/msys2-base-x86_64-20221216.sfx.exe'
```

I found the following on Stack overflow however this doesnt seem to do the trick. 
https://stackoverflow.com/questions/75018725/cant-update-ghc-on-windows

Really unsure what to do next. I have installed previously on other machines with the Haskell Platform when it was the main way to install with no issues so having to swap to this new experience and having things break is not the greatest

chreekat:
Does https://www.haskell.org/ghcup/guide/#certificate-authority-errors-curl look like it might be relevant?

hasufell:
Try: https://www.haskell.org/ghcup/guide/#certificate-authority-errors-curl

parkeradam:
New error now Which states the following 
```
Get-FileWCSynchronous 57.09 MB  00:00:06
Exception: Exec: Error executing command C:\WINDOWS\system32\taskkill.exe with arguments '/F /FI "MODULES eq msys-2.0.dll"'
```

Also curious why this isnt the default command if it's known to cause issues.

hasufell:
Seems like something is wrong on your machine and e.g. antivirus or windows defender is killing processes (msys2 installation) before it can finish.

parkeradam:
I've disable antivirus and I'm still getting the same error.

parkeradam:
![image](https://github.com/haskell/ghcup-hs/assets/30392071/f19b0e09-d901-44f7-a576-500b3d2708ba)

See the above screenshot for the error

[...]

PR #1094, opened 2024-07-05
Re #836 Improve documentation about curl and wget errors

mpilgrem:
Motivation: discussion at:
* #836


mpilgrem:
@hasufell, on the CI, something seems to be up with the macOS runners.

Issue #1101, opened 2024-07-10
keybindings in tutorial/advanced dialogs


hasufell:
In the comments of the issue it seems like it is solved. Currently we don't have keybindings for Tutorial not context menu. Also, part of the `KeyBinding` infrastructure in `ghcup` seems to replicate/duplicate that in [`brick`](https://hackage.haskell.org/package/brick-2.4/docs/Brick-Keybindings.html). Probably we should move it to brick.  

@dfordivam , do you have something in mind? are you already implementing something?

_Originally posted by @lsmor in https://github.com/haskell/ghcup-hs/issues/1099#issuecomment-2218194309_

dfordivam:
In order to support the same keybindings as used in the main navigation in the advanced install/compile popups, we would need to have a way for the user to edit the text input fields, while also allowing their custom keybindings to work for navigation actions like up, down, enter, quit (go back).

The idea I have in mind is to open an edit text field overlay, which has to be opened in order to begin editing. When the user has this opened then all the alpha-numeric + symbols keys would edit the text, and the only way to exit this would be via "Esc" or "Ctrl+c" key.

I already have an implementation which currently opens an overlay over the popup to allow selection from a list, like this


https://github.com/haskell/ghcup-hs/assets/681060/e2d189dc-46ce-4971-8c03-64762f89ac96

dfordivam:
This has been fixed and merged on master

PR #1102, opened 2024-07-13
Context Menu improvements; Select menu, Edit menu

dfordivam:
This includes a number of improvements like

- GHC(s) can be selected from a list of installed GHCs in bootstrap-ghc, hadrian-ghc, and HLS target-ghcs
- Build system can be selected from choices
- All the text edit inputs require opening an edit box by pressing "Enter".
  The help message an error message are also shown as the user types.

With this change it will be possible to use the user's keybindings in the advance popup's navigation. @lsmor 

tui-demo2


hasufell:
* bootstrap/hadrian GHC can be a full path to a binary
    * it can also be a ghc version in PATH that is not known to ghcup
* the titles of the popups are cut off (e.g. try the 'flavour' popup)

hasufell:
Another thing I noticed... it's very hard to distinguish required and optional settings.

Required settings should come first and maybe even be distinct from optional ones (not sure how exactly).

dfordivam:
> Another thing I noticed... it's very hard to distinguish required and optional settings.
> 
> Required settings should come first and maybe even be distinct from optional ones (not sure how exactly).

Ok, I have moved required fields to first. The simplest way to indicate the required fields is to mention them in the help message like this.

2024-07-19-121304_screenshot

> the titles of the popups are cut off (e.g. try the 'flavour' popup)

Fixed

> bootstrap/hadrian GHC can be a full path to a binary    it can also be a ghc version in PATH that is not known to ghcup

I have included additional fields to input them, as the UI gets considerably complex to support it via custom widgets.

[
2024-07-19-121841_screenshot
2024-07-19-121857_screenshot
](url)

hasufell:
> The simplest way to indicate the required fields is to mention them in the help message like this.

Most HTML UI uses an asterisk (`*` ...and possibly in red) and then have a line at the bottom: `* required fields`.

> I have included additional fields to input them, as the UI gets considerably complex to support it via custom widgets.

I'm not too excited about additional fields UX wise, because now it's not clear that those are "either or".

What does it take to implement a composite field that allows a drop-down + just a string?

dfordivam:
> I'm not too excited about additional fields UX wise, because now it's not clear that those are "either or".
> 
> What does it take to implement a composite field that allows a drop-down + just a string?

Ok makes sense. It is a bit complicated to do it the way other editable fields are handled, as the nesting of menu fields is tricky, but I did it the following style which avoids an additional overlay to edit the field's text. Does this look ok?

tui-demo3

hasufell:
> Does this look ok?

Yeah. I'd actually make the input field the first and is it possible to have a value for it when it's not selected, which could indicate that it is a custom input field?

dfordivam:
> Yeah. I'd actually make the input field the first and is it possible to have a "value" for it when it's not selected, which could indicate that it is a custom input field?

Done.

> Most HTML UI uses an asterisk (* ...and possibly in red) and then have a line at the bottom: * required fields.

In the case of compile GHC and HLS menus there is currently only one required field, so I think the current design where we  show an error in red if the field is empty is a fine UI.

tui-demo4

Also note that the gif I generated above doesn't show the help messages properly, but on my terminal I can see them quite clearly. Like the "Required fields: bootstrap-ghc" text.

2024-07-31-133950_screenshot

dfordivam:
And keeping the focus of the select menu's focusring on the first item after the text input field, is intentional.

Compare this 
2024-07-31-134738_screenshot

vs this, (if the focus is on text input field when the user opens the select menu)
2024-07-31-134801_screenshot

dfordivam:
I changed the help msg text to yellow, as it was not visible at all on the Ubuntu' gnome terminal' default theme.

Screenshot from 2024-08-07 13-51-59

Screenshot from 2024-08-07 13-54-51

dfordivam:
I have done more changes to address https://github.com/haskell/ghcup-hs/issues/1101

* Removed the hardcoded keybindings of Ctrl+C, Up, Down arrow keys in context menu, and tutorial navigation
* now the default quit keybinding of 'q' is used to move back / close context menu, 
* the up/down keys can be set to custom values like 'j' & 'k', and they should work in context menu navigation.
* The Editable input field is open in editing mode, it allows input of all keys including 'q', and it can only be closed by pressing 'Enter' (hardcoded)
* The custom text field of select field input is a bit tricky to handle. So I have decided with this approach: 
  The navigation keys like 'q', 'j', 'k' will work like normal to move focus among the different choices and select them by pressing Enter.
  But if the focus is on custom text field, then the only way to exit the edit mode is to press Enter, Up or Down keys.
  The help message in this scenario indicates that user must press Enter to finish editing like this

Screenshot from 2024-08-08 17-11-34

lib-tui/GHCup/Brick/App.hs
```diff
@@ -138,47 +145,36 @@ contextMenuHandler ev = do
...
+compileGHCHandler = menuWithOverlayHandler compileGHCMenu Actions.compileGHC CompileGHC.handler
 
 compileHLSHandler :: BrickEvent Name e -> EventM Name BrickState ()
-compileHLSHandler ev = do
-  ctx <- use compileHLSMenu
+compileHLSHandler = menuWithOverlayHandler compileHLSMenu Actions.compileHLS CompileHLS.handler
+
+-- | Passes all events to innerHandler if an overlay is opened
+-- else handles the exitKey and Enter key for the Menu's "OkButton"
+menuWithOverlayHandler
```
hasufell, line 158: I don't have a lot of formatting requirements, but `::` on the next line is a no-no, because it breaks grepping.

hasufell:
The PR is too big for me to review/digest in time, especially since I'm not familiar anymore with large parts of the TUI.

The functionality is fine afais.

lib-tui/GHCup/Brick/App.hs
dfordivam: fixed

PR #1104, opened 2024-07-13
Allow control of version for hls and stack during bootstrap

juergenpf:
Add support to control version of hls and stack installed during bootstrap


hasufell:
done and deployed

PR #1105, opened 2024-07-18
Windows bootstrap: Replace use of Exit with throw or return

dfordivam:
This works similarly with both `Invoke-Command -ScriptBlock` and `&`

I have tested most of the throw statements (by setting params, env vars, etc) + uninstall script changes.

Also contains a minor fix to cabal dir path check

The error is now printed like this, via `Write-Error`

2024-07-18-183518_screenshot


hasufell:
Is this good to merge or do you want to do more testing?

dfordivam:
I think its ok to merge

PR #1109, opened 2024-07-23
Make bootstrap-haskell.ps1 more transactional

### LINKED ISSUE #991: Windows installer script should be transactional
I tried the Windows Setup script at https://haskell.org/ghcup .
> Set-ExecutionPolicy Bypass -Scope Process -Force;[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; try { Invoke-Command -ScriptBlock ([ScriptBlock]::Create((Invoke-WebRequest https://www.haskell.org/ghcup/sh/bootstrap-haskell.ps1 -UseBasicParsing))) -ArgumentList $true } catch { Write-Error $_ }

It asked me questions with the option to "Abort".  I chose Abort after I already had answered "Reinstall" to GHCup.  As a consequence, the original installation had been backed up from `C:\ghcup` to a temporary directory and there was a new `C:\ghcup` which was empty except for a `bin` subfolder.

So, you cannot really "Abort" in the middle.  It is more like Hotel California:
> _You can check out at any time, but you can never leave._

I think such an installer should not interleave gathering information from the user and acting upon it.  
Instead, it should prepare a transaction and then commit it.  So, first gather the necessary information, then make a plan, and then execute the plan once it is clear that one can go through with it.
As it stands, a user having second thoughts and wanting to abort may end up in a broken state.  Moreover, as it installs itself to global locations like `C:\ghcup`, it will be broken for _all_ users on that machine.

hasufell:
Fixes #991

@andreasabel

### DIFF
--- a/scripts/bootstrap/bootstrap-haskell.ps1
+++ b/scripts/bootstrap/bootstrap-haskell.ps1
@@ -189,6 +189,8 @@ function Exec
     }
 }
 
+#### Gather system information and ask user questions ####
+
 # Only x86 32/64-bit is supported
 $SupportedArchitectures = 'AMD64', 'x86'
 if (!$SupportedArchitectures.contains($env:PROCESSOR_ARCHITECTURE)) {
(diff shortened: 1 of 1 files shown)


hasufell:
The diff is very hard to read and misleading. Better look at the full file. There's now two sections

Collecting information:

https://github.com/haskell/ghcup-hs/blob/9ab556c7931563c7015f649703d08994b500f9c4/scripts/bootstrap/bootstrap-haskell.ps1#L192-L515

Execute:

https://github.com/haskell/ghcup-hs/blob/9ab556c7931563c7015f649703d08994b500f9c4/scripts/bootstrap/bootstrap-haskell.ps1#L517-L711

dfordivam:
Looks good to me

PR #1117, opened 2024-08-11
Update stack.yaml, for Stack/macOS 13 CI

mpilgrem:
I've opened this pull request to see if I can fix the broken CI.


mpilgrem:
@hasufell, looks like it works: green ticks.

Issue #1118, opened 2024-08-12
Introduce pager mode (for `ghcup list`)


hasufell:
Following https://github.com/haskell/ghcup-hs/issues/1115 it appears the list is now big enough that it can be annoying to handle. Although users can just issue `ghcup list | less`, other tools such as `git` by default open a pager, unless the output is part of a pipe.

So I'm proposing the following behavior:

- new config option `pager: { list: bool, cmd: string }`
    - `cmd` is optional and can point to a full path or a filename (to be searched for in PATH)
    - when `pager: nil` -> paging disabled
    - when `pager: { list: false, ... }` -> paging disabled for list
    - `pager: true` is a shorthand for `pager: { list: true }` turning all boolean options `true` (if there are future ones)
    - `pager: string` is a shorthand for `pager: { list: true, cmd: string }`, turning all boolean options `true` (if there are future ones)
    - setting `cmd` to an empty string will disable paging as well
- new cli option `ghcup [-p|--paginate|-P|--no-pager]`
    -  overwrites config options
- lookup of pager in the given order
    - if `cmd` for `pager` config option is set, use it
    - otherwise check $GHCUP_PAGER env var
    - otherwise check $PAGER env var
    - otherwise check the following binaries in order
        - `most`
        - `more`
        - `less`
 - when to open the pager
     - default is off
     - if paging is enabled through `pager` config option
     - or if paging is enabled through cli option
 - when to not open pager (overwrites all other conditions)
     - if ghcup is run as part of a pipe (e.g. `ghcup list | grep foo` or `bar=$(ghcup list)`)
     - or if `ghcup list --raw-format` is passed
     - or if no valid pager is found
 - other remarks
     - warnings or anything else going to stderr is not shown in the pager
     - we may want pager for `ghcup --help` (compare with `git --help` vs `git -h`)
     - setting pager to `cat` should work too (and just print to stdout)

In the future we may:

1. switch the default for paging to `true` (need community input)
2. but only invoke a pager when the terminal height is less than the output (compare with `less -F`)

Useful libraries to check:

- https://hackage.haskell.org/package/terminal-size
- https://hackage.haskell.org/package/pager (I don't think I want to depend on this, but it can be an inspiration)

PR #1119, opened 2024-08-12
Implement pager support wrt #1118

hasufell:
TODO:

- [x] document
- [x] config options should be `pager`, not `pager-config` (adjust json)
- [x] deal with windows

### DIFF
--- a/app/ghcup/Main.hs
+++ b/app/ghcup/Main.hs
@@ -67,9 +67,8 @@ import qualified GHCup.Types                   as Types
 
 
 
-toSettings :: Maybe FilePath -> Options -> IO (Settings, KeyBindings, UserSettings)
-toSettings pagerCmd options = do
-  noColor <- isJust <$> lookupEnv "NO_COLOR"
+toSettings :: Bool -> Maybe FilePath -> Options -> IO (Settings, KeyBindings, UserSettings)
+toSettings noColor pagerCmd options = do
   userConf <- runE @'[ JSONError ] ghcupConfigFile >>= \case
     VRight r -> pure r
     VLeft (V (JSONDecodeError e)) -> do
@@ -77,10 +76,10 @@ toSettings pagerCmd options = do
       pure defaultUserSettings
     _ -> do
       die "Unexpected error!"
-  pure $ (\(s', k) -> (s', k, userConf)) $ mergeConf options userConf noColor
+  pure $ (\(s', k) -> (s', k, userConf)) $ mergeConf options userConf
  where
-   mergeConf :: Options -> UserSettings -> Bool -> (Settings, KeyBindings)
-   mergeConf Options{..} UserSettings{..} noColor =
+   mergeConf :: Options -> UserSettings -> (Settings, KeyBindings)
+   mergeConf Options{..} UserSettings{..} =
      let cache       = fromMaybe (fromMaybe (Types.cache defaultSettings) uCache) optCache
          metaCache   = fromMaybe (fromMaybe (Types.metaCache defaultSettings) uMetaCache) optMetaCache
          metaMode    = fromMaybe (fromMaybe (Types.metaMode defaultSettings) uMetaMode) optMetaMode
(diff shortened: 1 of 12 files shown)


app/ghcup/Main.hs
```diff
@@ -68,17 +69,18 @@ import qualified GHCup.Types                   as Types
 toSettings :: Options -> IO (Settings, KeyBindings, UserSettings)
 toSettings options = do
   noColor <- isJust <$> lookupEnv "NO_COLOR"
+  pagerCmd <- getPager
```
hasufell, line 72: We're unconditionally executing this, which is a bit unfortunate... I think lazy IO could be fine here.

hasufell:
When using `less`... the unicode stuff doesn't work very well. You have to either export `LESSCHARSET=utf-8` or use `PAGER='less -r'`.

lib/GHCup/Utils/Pager.hs
```diff
@@ -0,0 +1,54 @@
...
+        Just r  -> pure $ Just r
+        Nothing -> lookupEnv "PAGER" >>= \case
+          Just r' -> pure $ Just r'
+          Nothing -> do
+                findExecutable "most"
+            <|> findExecutable "more"
+            <|> findExecutable "less"
+
+sendToPager :: FilePath -> Text -> IO ()
+sendToPager pager text = sendToPagerConduit pager (sourceLbs (BSL.fromStrict . E.encodeUtf8 $ text))
```
hasufell, line 37: Assuming UTF-8 here is a bit unfortunate. But I don't know if ghcup behaves well wrt unicode symbols if you have esoteric encodings set anyway.

lib/GHCup/Utils/Pager.hs
```diff
@@ -0,0 +1,54 @@
...
+
+sendToPagerConduit :: FilePath -> ConduitT () ByteString (ResourceT IO) () -> IO ()
+sendToPagerConduit pager producer = do
+     ((Just stdinH),_,(Just stderrH),ph) <-
+       createProcess
```
dfordivam, line 42: I am not confident about this, for a start using `withCreateProcess` might be better. And should the std_out be `Inherit`, and delegate_ctlc = True?

I had looked at the implementation done by git and it is doing signal handling changes, like ignoring signals which ought to be handled by pager.

dfordivam:
For windows the `more` might work, will need testing i it work from within the process's invocation, writing to stdin (https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/more)

hasufell:
> For windows the `more` might work, will need testing i it work from within the process's invocation, writing to stdin (https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/more)

in powershell, it does not... in msys2 anything works

dfordivam:
> When using `less`... the unicode stuff doesn't work very well. You have to either export `LESSCHARSET=utf-8` or use `PAGER='less -r'`.

I had `LESS=-R` in my config and hence didn't notice this issue.

dfordivam:
LGTM, the incremental improvements could be done later.

dfordivam (approved):

lib/GHCup/Utils/Pager.hs
```diff
@@ -0,0 +1,66 @@
...
+    $ \(Just stdinH) _ (Just stderrH) ph -> do
+        forM_ text $ T.hPutStrLn stdinH
+        hClose stdinH
+        exitCode <- waitForProcess ph
+        case exitCode of
+          ExitFailure i ->
+            do errContents <- hGetContents stderrH
+               fail (unlines [mappend "Pager exited with exit code " (show i)
+                             ,errContents])
+          _ -> pure ()
```
dfordivam, line 51: I looked into the `more` support, by looking at its source code. It reads from the STDERR and thus gets stuck if we create a new pipe for std_err. I confirmed that this change fixes `more` on both linux and windows
 https://github.com/util-linux/util-linux/blob/81ea117e0396274808a77d40bf11af44ea047285/text-utils/more.c#L804C11-L804C24

```suggestion
-- 'more' reads from STDERR, and requires std_err to be 'Inherit'
sendToPager :: FilePath -> [Text] -> IO (Either IOException ())
sendToPager pager text = try @IOException
    $ withCreateProcess (shell pager) { std_in = CreatePipe
                                      , std_err = Inherit
                                      , std_out = Inherit
                                      , delegate_ctlc = True
                                      }
    $ \(Just stdinH) _ _ ph -> do
        forM_ text $ T.hPutStrLn stdinH
        hClose stdinH
        exitCode <- waitForProcess ph
        case exitCode of
          ExitFailure i -> fail ("Pager exited with exit code " <> show i)
          _ -> pure ()
```

hasufell: Great find

PR #1123, opened 2024-08-25
Switch to variant package

hasufell:


hseg:
Presumably this section in https://github.com/haskell/ghcup-hs/blob/master/docs/dev.md?plain=1#L39 should be changed:
```
### Using [Excepts](https://hackage.haskell.org/package/haskus-utils-variant-3.0/docs/Haskus-Utils-Variant-Excepts.html) as a beefed up ExceptT

This is an open variant, similar to [plucky](https://hackage.haskell.org/package/plucky) or [oops](https://github.com/i-am-tom/oops) and allows us to combine different error types. Maybe it is too much and it's a little bit [unergonomic](https://github.com/haskus/packages/issues/32) at times. If it really hurts maintenance, it will be removed. It was more of an experiment.
```

PR #1126, opened 2024-09-03
stack.yaml: Remove spurious flags

hseg:
b9adc1f accidentally added `win32-2-13-1: true` to `ansi-terminal` as well. Stack complains that no such flag exists, preventing building.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -72,9 +72,6 @@ flags:
   mintty:
     win32-2-13-1: true
 
-  ansi-terminal:
-    win32-2-13-1: true
-
 ghc-options:
   "$locals": -O2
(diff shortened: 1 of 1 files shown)


hasufell:
https://github.com/haskell/ghcup-hs/actions/runs/10686787693/job/29631795298#step:3:1

hseg:
Note that the run failure there is due to an incorrect `Win32` version, not this flag. Pushing a fix, one moment (I don't have access to a windows machine I can test this on, so please bear with my using the CI to check).

hseg:
Oh, wait. I must've looked at the hackage page for the wrong version, because as of `ansi-terminal 0.11.5` (which is the version ghcup uses), the flag still exists.
I don't know what caused my own setup to choke on the flag, but clearly the original commit is in the right. Sorry for the noise.

PR #1127, opened 2024-09-11
Support GHC-9.6 (rebased)

hseg:
Rebased version of #997 onto `master`.

Notable edits:
- `app/ghcup/Main`:
  - Update `Cabal.Plan` -> `PlanJson` -> `GHCup.PlanJson`
- `lib-opt/GHCup/OptParse/Common.hs`: Reflect extraction of some parts to
  `lib/GHCup/Utils/Parsers.hs`
- Breakup of `app/ghcup/BrickMain.hs`
- freezefiles:
  - Picked up most changes from fix-hashable-mess, though see
    5b41f7970d2480332e1da7905416e72117122233 for doubts regarding my choices.
    Similarly, picked up some version bumps on the way (the unix freeze files
    are more affected by this).
    I'm least confident about this edit, it might require polishing
  - Also picked up `haskus-*` -> `variant`
  - UPDATE: [Ended up regenerating these for unix, windows generation TBA once the other builds succeed](https://github.com/haskell/ghcup-hs/pull/1127#issuecomment-2343686750)
- `cabal.*.project`:
  - Given the commit for the `tar` repository that had been added is now
    released, I dropped the `source-repository-package` directive
  - Picked up the `hashable` needing `-arch-native`
  - UPDATE: [Removed stale traces of `cabal-{install-parsers,plan}`](https://github.com/haskell/ghcup-hs/pull/1127#issuecomment-2343647268)
- `cabal.project.release`:
  - UPDATE: [Removed the pinned `tar` commit](https://github.com/haskell/ghcup-hs/pull/1127#issuecomment-2343587594), like in the other `cabal.*.project` files
  - UPDATE: [Removed stale traces of `cabal-{install-parsers,plan}`](https://github.com/haskell/ghcup-hs/pull/1127#issuecomment-2343647268)
  - UPDATE: [Removed stale traces of `streamly` and co](https://github.com/haskell/ghcup-hs/pull/1127#issuecomment-2343849068)
  - UPDATE: [Reorganized to follow the same order as the other `cabal.*.project` files](https://github.com/haskell/ghcup-hs/pull/1127#issuecomment-2343856529)
  - UPDATE: [Added the conditional `ghcup +tar` logic found in other `cabal.*.project` files](https://github.com/haskell/ghcup-hs/pull/1127#issuecomment-2343856529)
- `ghcup.cabal`:
  - Dropping `cabal-{install-parsers,plan}` in favor of `Cabal-syntax` and
    creation of `GHCup.{CabalConfig,PlanJson,Utils.Parsers}`
  - Addition of `parsec`, `utf8-string`
  - `haskus-*` -> `variant`
  - `optics` moving into `app-common-depends`
  - Extraction of `ghcup-tui`
  - Removal of `includes:`
- `stack.yaml`:
  - `haskus-*` -> `variant`
  - Use `tar-0.6.3.0` instead of commit, pick up `ghcup +tar` flag
  - Add `mintty +win32-2-13-1` flag
- Bumped `tested-with` to `ghc-9.6.5` based on the existence of `cabal.ghc965.*`
- macOS version bumped to 12
- Deleted `.orig` files after it was confirmed they were unintentionally added.

Closes: #979, #997


hseg:
I don't know Github workflows well, it seems `.github/release.yaml` and `.github/workflows/shimgen.yaml` should be upgraded away from `upload-artifact@v2`? https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

hseg:
Ah wait, perhaps the remaining reference to the pinned commit of https://github.com/haskell/tar in `cabal.project.release` should be removed?

hseg:
(note that the reference to a pinned commit in `uri-bytestring` is the as-yet-unmerged https://github.com/Soostone/uri-bytestring/pull/64, so it can't be removed yet)

hseg:
Hrm. I should probably not be force-pushing commits anymore now you're reviewing them, right? Sorry for that, will stop doing so if there are more modifications to be made

hseg:
BTW, it is unclear to me why these lines were removed from `cabal.project.release` in the original branch:
```
elif os(mingw32)
  constraints: zlib +bundled-c-zlib,
               lzma +static,
               text -simdutf,
               vty-windows >=0.2.0.2
  if impl(ghc >= 9.4)
    constraints: language-c >= 0.9.3
```
(Now, tbf they do occur in `cabal.project`, but `mingw32` is the only platform that was previously duplicated there, and it is unclear why it was deduplicated in favor of `cabal.project` rather than `cabal.project.release` like `linux`, `darwin` and `freebsd`)

hseg:
Wait, shouldn't I have dropped references to ` cabal-{install-parsers,plan}`, given the migration away from it?

hseg:
Oh wait, missed the existence of `scripts/dev/refreeze.sh`.
Should I be using that, instead? I don't have access to a windows box, so I can't create the windows freezefiles.

hasufell:
I can handle the windows files. just update the unix ones

hseg:
Missed one more migration: `streamly` -> `conduit` in `cabal.project.release`

hseg:
Was 411ac8dc3118a1863dcd82a2eca8d39db66a200f supposed to affect `cabal.project.release` as well?
It guarded `ghcup +tar` behind an `impl(ghc < 9.0)`

.github/workflows/cabal.project.yaml
```diff
@@ -20,12 +20,14 @@ jobs:
       fail-fast: false
       matrix:
         os: [macOS-12, windows-latest, ubuntu-latest]
-        ghc: ["8.10.7", "9.0.2", "9.2.8", "9.4.8"]
+        ghc: ["8.10.7", "9.0.2", "9.2.8", "9.4.8", "9.6.5"]
```
hasufell, line 23: Why not 9.6.6?

ghcup.cabal
```diff
@@ -65,44 +66,44 @@ flag tar
...
-    , base                   >=4.12     && <5
-    , bytestring             >=0.10     && <0.12
-    , containers             ^>=0.6
-    , deepseq                ^>=1.4
-    , directory              ^>=1.3.6.0
-    , filepath               >=1.4.101.0
-    , variant                ^>=1.0
-    , megaparsec             >=8.0.0    && <9.3
-    , mtl                    ^>=2.2
+    , aeson                 >=1.4
```
hasufell, line 69: what's the point of reformatting here?

cabal.project.release
```diff
@@ -1,11 +1,23 @@
-packages: ./ghcup.cabal
-
-optional-packages: ./vendored/*/*.cabal
+import: cabal.project
 
 optimization: 2
 
-package ghcup
-  flags: +tui -tar
+if impl(ghc < 9.0)
```
hasufell, line 5: This is not correct. The release project file always wants to build with `-tar`. The tar flag is only there to circumvent complicated errors during development, which sometimes happens due to libarchive.

cabal.project.release
```diff
@@ -1,11 +1,23 @@
...
+    package ghcup
+        flags: +tui -tar
+else
+    package ghcup
+        flags: +tui +tar
+
+constraints: http-io-streams -brotli,
```
hasufell, line 12: Why add these when you already did `import: cabal.project`?

.github/workflows/cabal.project.yaml
hseg: Old branch was on 9.6.5, wasn't sure if you'd appreciate me bumping it.

ghcup.cabal
hseg: I'd noticed the spacing had ended up inconsistent after the last couple of edits there, and just blasted it with `cabal-fmt`. I can either drop this reformatting, take the less-noisy option of just realigning the outliers, or reformat it with [gild](https://taylor.fausak.me/2024/02/17/gild/) to avoid such issues in the future. Your choice. (Also, please let me know if you'd prefer me just dropping/squashing the reformatting commit or if you'd like me to revert the reformatting on top of the current tip)

cabal.project.release
hseg: Ah, OK. I'll drop a note in a commit reverting this, let me know if you want me to squash them together (so the resulting commit will just add a comment noting this discrepancy vs the other cabal.project files)

cabal.project.release
hseg: True, though by that logic, shouldn't `cabal.project.release` be simplified to:
```cabal
import: cabal.project

optimization: 2

if os(linux)
  if arch(x86_64) || arch(i386)
    package *
      ghc-options: -split-sections -optl-static
elif os(darwin)
  constraints: zlib +bundled-c-zlib,
               lzma +static
elif os(freebsd)
  constraints: zlib +bundled-c-zlib,
               zip +disable-zstd
  package *
    ghc-options: -split-sections
```

cabal.project.release
hseg: Also, is this why the `mingw32` flags are present in `cabal.project` (and will be imported in `cabal.project.release`, whereas the `linux`, `darwin`, and `freebsd` flags are only present in `cabal.project.release`? Windows causes more trouble that needs to be caught at dev time, but the others are mostly just optimizations for release time?

cabal.project.release
hseg: BTW, these constraints already existed at the bottom of the file -- I just moved them to the same place they were in the other `cabal.project` files

ghcup.cabal
hseg: Just to make things concrete, I pushed a branch to my fork with the `gild`-formatted version of the cabal files. Let me know if it's in any way useful to you
https://github.com/hseg/ghcup-hs/tree/reformat-cabal

hasufell: the reformatting makes it painful to fix merge conflicts

hasufell:
i squashed and rebased and set you as co-author in the commit, thanks: https://github.com/haskell/ghcup-hs/pull/1132

hseg:
Thanks, have had much less free time in the past week, so I couldn't get back to you.

PR #1133, opened 2024-09-29
Add multi-user install support in windows bootstrap

dfordivam:
By setting up the env vars at the "Machine" level.

If there are other users currently logged in, they will have to sign out and login again for the global env vars to take effect.


dfordivam:
The cabal config adjustments didn't happen with this. Those will have to be done manually by each user.

hasufell:
Hmm... do they have to run the script as administrator or will windows create a popup for privilege escalation?

dfordivam:
So this enhancement is a relatively minor convenience over manual env vars adjustments. And changing the system wide PATH via this script with admin privilege is also slightly scary.
I suggest we ditch this idea and only provide instructions in guide for manual setup of env vars.

hasufell:
Agreed

PR #1151, opened 2024-11-12
Allow specifying a custom version in advance install menu

dfordivam:
This field is necessary as the user may be installing some arbitrary version using a URI, and there is no way to specify a version for that, (like the command line `ghcup install -u <uri> <version>` supports.)

The `GHCTargetVersion` used for `instVersion` is not ideal for handling installation of other tools, but since we have a single AdvanceInstall menu shared for all tools, I think its ok to use this to avoid a bigger change to the code.


lsmor:
I've been a little bit out of FOSS lately but I have sometime now. Let me get use to the new changes an I'll make a review asap.

hasufell:
@dfordivam can you rebase?

dfordivam:
@hasufell rebase done

PR #1152, opened 2024-11-14
On windows show 'Y' in checkbox, instead of 'I'

dfordivam:


hasufell:
FreeBSD failure is expected

lib-tui/GHCup/Brick/Common.hs
```diff
@@ -179,6 +180,13 @@ notInstalledSign = "X "
 notInstalledSign = "✗ "
 #endif
 
+checkBoxSelectedSign :: String
+#if IS_WINDOWS
```
hasufell, line 184: We already have a function for that: `isWindows`. We don't need the explicit CPP everywhere.

dfordivam: Fixed

hasufell (approved):

PR #1154, opened 2024-11-15
Switch to xz

hasufell:
Following https://github.com/hasufell/lzma-static/pull/11

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,4 +1,4 @@
-packages: ./ghcup.cabal
+import: cabal.project.common
 
 optional-packages: ./vendored/*/*.cabal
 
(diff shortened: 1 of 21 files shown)


cabal.project
```diff
@@ -23,7 +23,6 @@ constraints: text -simdutf
 
 if os(mingw32)
   constraints: zlib +bundled-c-zlib,
-               lzma +static,
                text -simdutf,
```
dfordivam, line 26: `xz -system-xz` isn't necessary here too?

dfordivam (approved):
Looks fine, though I didn't understand mingw32 constraints

cabal.project
hasufell: No. The `cabal.project` is for development. Linking to system libs if they exist is fine.

Issue #1050, opened 2024-04-29
Easier selection of known release channels


hasufell:
1. ask during bootstrap script which release channels should be activated
2. maybe have aliases for known release channels, so that the user can do `ghcup config add-release-channel prereleases`
    - where should the mapping exist? In ghcup code, another yaml or in one of the "base" yamls?
3. Support selecting/managing release channels via the TUI

@lsmor

lsmor:
About the first two points.... I got no idea. About this:

> 3\. Support selecting/managing release channels via the TUI

Surely we can a new button in the contex menu, with "Add release channel" and something like this

```
|--------------- Add New Channel -----------------|
|                                                 |
| From known Channels: ◦ nightlies                |
|                      ◉ prereleases              |
|                      ◦ latest                   |
|                                                 |
| From URL:  [ Enter your URL here           ]    |
|---------------- Add New Channel ----------------|
```

dfordivam:
> * ask during bootstrap script which release channels should be activated


In interactive mode we can add support for this, in both unix and win.

```
		warn "Do you want to enable other release channels?"
		warn "The following release channels can be enabled during installation"
		warn "cross, nightlies, prereleases, and vanilla"
		warn ""
		warn "[Y] Yes  [N] No  [?] Help (default is \"N\")."
		warn ""
```
Then loop and ask for each channel Y/N, giving brief info about each and their URL.
We can limit this to cross, prereleases, and vanilla only (ie not include nightlies here)

> * maybe have aliases for known release channels, so that the user can do `ghcup config add-release-channel prereleases`
>   
>   * where should the mapping exist? In ghcup code, another yaml or in one of the "base" yamls?

Aliases can be added in ghcup code

> * Support selecting/managing release channels via the TUI

We can add a Context Menu for GHCup, which allows various config related changes, including channel selection.

The context menu can be opened by moving focus on any "GHCup" row and pressing "Enter"
In TUI the GHCup row is always shown, even in case of an empty metadata file.

Currently "Install" menu is shown in GHCup, we can add more things to it gradually.

There should be checkboxes for each of the channels; on "Save" the config will be updated.
There could be an additional Text input field (via popup) which allows adding custom URLs
```
 ┌─────────────────────────────────────────Context Menu for GHCup 0.1.30.0────────────────────────────────────────┐
 │┌───────┐                                                                                                       │
 ││Install│        Advance Installation Settings                                                                  │
 │└───────┘                                                                                                       │
[...]

PR #1155, opened 2024-11-15
Support aliases in 'config add-release-channel' and '--url-source'

dfordivam:
Add the URIs of known release channels in code, and allow choosing them via aliases.

Ref: https://github.com/haskell/ghcup-hs/issues/1050 

With this its possible to just do 🙃
`ghcup -s vanilla install ghc 9.8.3`


lib/GHCup/Types.hs
```diff
@@ -394,6 +394,13 @@ data NewURLSource = NewGHCupURL
 
 instance NFData NewURLSource
 
+-- | Alias for ease of URLSource selection
+data ChannelAlias = MainChannel
+                  | CrossChannel
+                  | PrereleasesChannel
+                  | VanillaChannel
```
hasufell, line 401: This is now a bit confusing, because the config supports the value `GHCupURL` and `StackSetupURL`, but none of the other channels.

https://github.com/haskell/ghcup-hs/blob/a32a5597300b777b488e1649b1a662dc5cce150f/data/config.yaml#L54-L92

https://github.com/haskell/ghcup-hs/blob/a32a5597300b777b488e1649b1a662dc5cce150f/lib/GHCup/Types.hs#L388-L393

And the json instances are finicky.

hasufell: It creates inconsistency, because `NewURLSource` is valid in both the config and in the cli: `ghcup -s GHCupURL tui`.

Now we have another set of aliases that don't work in the config but in the cli. I have tried to keep both of them in sync as much as possible.

dfordivam: Now the alias is part of `URLSource` & `NewURLSource`

lib-opt/GHCup/OptParse/Config.hs
```diff
@@ -96,8 +98,10 @@ configFooter = [s|Examples:
   ghcup config init
 
   # set <key> <value> configuration pair
-  ghcup config set <key> <value>|]
+  ghcup config set <key> <value>
 
+  # add a release channel
+  ghcup config add-release-channel vanilla|]
```
hasufell, line 104: Vanilla isn't a good example, because we usually want to set it instead of adding it.

So I think we want twe examples:

- adding prereleases
- setting vanilla (discrding everything else)

lib/GHCup/Types.hs
```diff
@@ -379,6 +379,7 @@ instance Pretty TarDir where
 -- | Where to fetch GHCupDownloads from.
 data URLSource = GHCupURL
                | StackSetupURL
+               | ChannelAlias  ChannelAlias
```
hasufell, line 382: We don't want to touch `URLSource`. It's there for backwards comp and already supports `NewURLSource` through the `SimpleList` constructor.

dfordivam: Ok, fixed

lib-opt/GHCup/OptParse/Config.hs
dfordivam: fixed

lib/GHCup/Types.hs
dfordivam: hmm, the `ghcup -s vanilla install ghc 9.8.3` no longer works without `URLSource`. I should probably convert `optUrlSource   :: Maybe URLSource` to `optUrlSource   :: Maybe NewURLSource`

hasufell: > I should probably convert optUrlSource   :: Maybe URLSource to optUrlSource   :: Maybe NewURLSource

That sounds like a breaking change, no?

Did you check `parseUrlSource` in `GHCup.Utils.Parsers`?

dfordivam: I did this in backward compatible way, by modifying `parseUrlSource` too. It now handles both `URLSource` and `NewURLSource`

`cabal run exe:ghcup -- -s "{\"OwnSource\": "\"https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-prereleases-0.0.8.yaml\""}" list`

`cabal run exe:ghcup -- -s vanilla install ghc 8.10.2`

lib/GHCup/Types.hs
dfordivam: Correction alias is only part of `NewURLSource`

PR #1162, opened 2024-11-22
Allow to mark a single bindist as experimental

hasufell:
As discussed with @bgamari @AndreasPK

Screenshot_2024-11-22_23-40-59


dfordivam:
Oh, so the `Tag` is for a particular bindist (for a arch + platform + os), instead of a version as a whole. IIUC this is for distributing bindist for platforms not fully supported like *BSD, etc.

dfordivam (approved):
LGTM

PR #1172, opened 2024-12-09
Don't ignore ghc compile bootstrap error, fix --remote-source-dist

dfordivam:
Also included in this
- support of --patch with --remote-source-dist
- support compilation of bootstrapped src tar with --remote-source-dist
  This does not contain boot file, and currently it results in following 
```
[ Info  ] Unpacking: ghc-9.8.4-src.tar.xz to /home/divam/.ghcup/tmp/ghcup-7fbd043d844b981c
ghcup: user error (Pattern match failure in 'do' block at lib/GHCup/GHC.hs:874:11-20)
```

### DIFF
--- a/lib/GHCup/GHC.hs
+++ b/lib/GHCup/GHC.hs
@@ -868,8 +868,11 @@ compileGHC targetGhc crossTarget vps bstrap hghc jobs mbuildConfig patches aargs
         tmpDownload <- lift withGHCupTmpDir
         tmpUnpack <- lift mkGhcupTmpDir
         tar <- liftE $ download uri Nothing Nothing Nothing (fromGHCupPath tmpDownload) Nothing False
-        (bf, tver) <- liftE $ cleanUpOnError @'[UnknownArchive, ArchiveResult, ProcessError] tmpUnpack $ do
+        (bf, tver) <- liftE $ cleanUpOnError @'[UnknownArchive, ArchiveResult, ProcessError, PatchFailed, DownloadFailed, DigestError, ContentLengthError, GPGError] tmpUnpack $ do
           liftE $ unpackToDir (fromGHCupPath tmpUnpack) tar
+
+          liftE $ applyAnyPatch patches (fromGHCupPath tmpUnpack)
+
           let regex = [s|^(.*/)*boot$|] :: B.ByteString
           [bootFile] <- liftIO $ findFilesDeep
             tmpUnpack
--- a/lib/GHCup/GHC.hs
+++ b/lib/GHCup/GHC.hs
@@ -868,23 +868,22 @@ compileGHC targetGhc crossTarget vps bstrap hghc jobs mbuildConfig patches aargs
         tmpDownload <- lift withGHCupTmpDir
         tmpUnpack <- lift mkGhcupTmpDir
         tar <- liftE $ download uri Nothing Nothing Nothing (fromGHCupPath tmpDownload) Nothing False
-        (bf, tver) <- liftE $ cleanUpOnError @'[UnknownArchive, ArchiveResult, ProcessError, PatchFailed, DownloadFailed, DigestError, ContentLengthError, GPGError] tmpUnpack $ do
+        (workdir, tver) <- liftE $ cleanUpOnError @'[UnknownArchive, ArchiveResult, ProcessError, PatchFailed, DownloadFailed, DigestError, ContentLengthError, GPGError] tmpUnpack $ do
           liftE $ unpackToDir (fromGHCupPath tmpUnpack) tar
 
-          liftE $ applyAnyPatch patches (fromGHCupPath tmpUnpack)
-
-          let regex = [s|^(.*/)*boot$|] :: B.ByteString
-          [bootFile] <- liftIO $ findFilesDeep
+          let regex = [s|^(.*/)*compiler/ghc.cabal.in$|] :: B.ByteString
+          (ghcCabalIn:_) <- liftIO $ findFilesDeep
             tmpUnpack
-            (makeRegexOpts compExtended
-                           execBlank
-                           regex
-            )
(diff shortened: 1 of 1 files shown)


lib/GHCup/GHC.hs
```diff
@@ -868,20 +868,25 @@ compileGHC targetGhc crossTarget vps bstrap hghc jobs mbuildConfig patches aargs
         tmpDownload <- lift withGHCupTmpDir
         tmpUnpack <- lift mkGhcupTmpDir
         tar <- liftE $ download uri Nothing Nothing Nothing (fromGHCupPath tmpDownload) Nothing False
-        (bf, tver) <- liftE $ cleanUpOnError @'[UnknownArchive, ArchiveResult, ProcessError] tmpUnpack $ do
+        (workdir, tver) <- liftE $ cleanUpOnError @'[UnknownArchive, ArchiveResult, ProcessError, PatchFailed, DownloadFailed, DigestError, ContentLengthError, GPGError, NotFoundInPATH] tmpUnpack $ do
           liftE $ unpackToDir (fromGHCupPath tmpUnpack) tar
-          let regex = [s|^(.*/)*boot$|] :: B.ByteString
-          [bootFile] <- liftIO $ findFilesDeep
+
+          let regex = [s|^(.*/)*compiler/ghc.cabal.in$|] :: B.ByteString
```
hasufell, line 874: can you explain this?

dfordivam: This change is to fix the `workdir` retrieval. The bootstrapped src does not contain `boot` file.

finding the directory name using `ghc.cabal.in` seems to be better.

lib/GHCup/GHC.hs
```diff
@@ -1037,6 +1042,24 @@ compileGHC targetGhc crossTarget vps bstrap hghc jobs mbuildConfig patches aargs
...
+                      , MonadThrow m
+                      )
+                   => GHCupPath
+                   -> Excepts '[ProcessError, NotFoundInPATH] m ()
+  bootAndConfigure tmpUnpack = do
+    let bootFile = fromGHCupPath tmpUnpack </> "boot"
+    hasBootFile <- liftIO $ doesFileExist bootFile
+    when hasBootFile $ do
+      lift $ logDebug "Doing ghc-bootstrap"
+      python3 <- liftE $ makeAbsolute "python3"
```
hasufell, line 1059: This is not necessary. The reason we swallow the bootstrap error is because of:

```hs
          tver <- liftE $ catchAllE @_ @'[ProcessError, ParseError, NotFoundInPATH] @'[] (\_ -> pure Nothing) $ fmap Just $ getGHCVer
```

`execLogged` already returns the correct `Left` when `executeFile` can't find the exe. So separating bootstrap from retrieving the GHC version should be enough.

lib/GHCup/GHC.hs
```diff
@@ -1037,6 +1042,24 @@ compileGHC targetGhc crossTarget vps bstrap hghc jobs mbuildConfig patches aargs
...
+      lEM $ execLogged python3 ["./boot"] (Just $ fromGHCupPath tmpUnpack) "ghc-bootstrap" Nothing
+      liftE $ configureWithGhcBoot Nothing [] (Just $ fromGHCupPath tmpUnpack) "ghc-bootstrap"
```
hasufell, line 1061: I'm not sure this belong outside of `getGHCVer`. The reason we run configure here (without any meaningful arguments) is because it creates the VERSION file, which we then use. So this is not the "proper" configure. The only reason we do this is to get the GHC version.

See GHCs `configure.ac`:

```
dnl Create the VERSION file, satisfying #22322.
printf "$ProjectVersion" > VERSION
```

lib/GHCup/GHC.hs
hasufell: I'm not sure this file will always exist in that place. If there is no `boot` file, then we can assume there is a `configure` file. And the topmost such configure file is what we need. We should depend on the files we actually use.

From what I understand `findFilesDeep` is breadth first, so we can just take the first find (possibly using the conduit functions to cut off early).

lib/GHCup/GHC.hs
dfordivam: Ok, in that case it does belong here, as the bootstrapped ghc will contain the VERSION file.

lib/GHCup/GHC.hs
dfordivam: I reverted back to `boot` file, the file gets renamed now in hadrian when making release builds.

```
      -- And move ./boot so we can't accidentally call it in CI                                                           
      moveFile (dest -/- "boot") (dest -/- "boot.source")       
```

lib/GHCup/GHC.hs
dfordivam: reverted this

lib/GHCup/GHC.hs
```diff
@@ -1035,6 +1043,24 @@ compileGHC targetGhc crossTarget vps bstrap hghc jobs mbuildConfig patches aargs
...
+                      , HasLog env
+                      , MonadIO m
+                      , MonadThrow m
+                      )
+                   => GHCupPath
+                   -> Excepts '[ProcessError, NotFoundInPATH] m ()
+  bootAndGenVersion tmpUnpack = do
+    let bootFile = fromGHCupPath tmpUnpack </> "boot"
+    hasBootFile <- liftIO $ doesFileExist bootFile
+    when hasBootFile $ do
```
hasufell, line 1058: I'm not sure. As in: if we have booted, does it mean there is always a `VERSION` file? From what I understand one has to run configure too.

@bgamari @andreasPK

dfordivam: The `boot` file is being renamed to `boot.source` when doing `hadrian build source-dist`, so this is a scenario different from simply bootstrapped ghc. And as part of this `build source-dist` the `VERSION` is always being copied. 

ref: https://gitlab.haskell.org/ghc/ghc/-/blob/master/hadrian/src/Rules/SourceDist.hs#L123-143

PR #1196, opened 2025-01-02
Add 'default' alias

hasufell:


lib/GHCup/Version.hs
```diff
@@ -89,6 +89,8 @@ pvpFromList = V.PVP . NE.fromList . fmap fromIntegral
 
 channelURL :: ChannelAlias -> URI
 channelURL = \case
+  DefaultChannel -> ghcupURL
+  StackChannel -> stackSetupURL
```
dfordivam, line 93: This will fail here

`dl' (NewChannelAlias c) = fmap Left $ liftE (getBase $ channelURL c) >>= liftE . decodeMetadata @GHCupInfo` 

So add this?

`  dl' (NewChannelAlias StackChannel) = fmap Right $ liftE (getBase stackSetupURL) >>= liftE . decodeMetadata @Stack.SetupInfo` ?

dfordivam (approved):

Issue #1213, opened 2025-01-15
Speed up metadata parsing


hasufell:
It appears that `parseURI` takes a lot of time.

Use https://downloads.haskell.org/~ghcup/tmp/ghcup.eventlog.json to load into https://www.speedscope.app/

![Image](https://github.com/user-attachments/assets/0df982b6-9df9-4d45-82a8-b1b5285d66f3)

hasufell:
https://github.com/haskell/ghcup-hs/pull/1214

hasufell:
Old:

```
$ time ghcup -s https://ghc.gitlab.haskell.org/ghcup-metadata/ghcup-nightlies-2024-0.0.7.yaml list --show-nightly --since 2024-12-12 -t ghc -r
...

________________________________________________________
Executed in  370.73 millis    fish           external
   usr time  320.08 millis    0.00 micros  320.08 millis
   sys time   50.90 millis  884.00 micros   50.01 millis
```

New:

```
$ time ghcup -s https://ghc.gitlab.haskell.org/ghcup-metadata/ghcup-nightlies-2024-0.0.7.yaml list --show-nightly --since 2024-12-12 -t ghc -r
...

________________________________________________________
Executed in  147.32 millis    fish           external
   usr time  130.23 millis    0.28 millis  129.95 millis
   sys time   11.89 millis    1.06 millis   10.83 millis
```

PR #1214, opened 2025-01-15
Avoid repeated calls to 'hlsGHCVersions'

hasufell:
Related to #1213


hasufell:
If we merge this, when we'll have to adjust ghcup-metadata to check for well-formedness of the URLs.

dfordivam (approved):
That sounds ok, it does save a lot on parse on every execution of ghcup.

hasufell:
I got 20ms more out of it by switching back to `yaml-streamly`, but I'm not sure I want to go there.

PR #1229, opened 2025-02-09
Fix handling of 'ghc' subdir in parser for obtaining set ghc version

### LINKED ISSUE #1215: ghcup set fails on windows
From `ghcup tui`, trying to set the ghc version fails:

```
[ Info  ] This Version of GHC you are trying to set is not installed.
[ ...   ] Would you like to install it first? [Y/N]:
y
[ Info  ] Setting now...
[ Info  ] This Version of GHC you are trying to set is not installed.
[ ...   ] Would you like to install it first? [Y/N]:
y
[ Info  ] Setting now...
[ Info  ] This Version of GHC you are trying to set is not installed.
[ ...   ] Would you like to install it first? [Y/N]:
n
Error: [GHCup-00130] The version 'ghcup' of the tool ghc is not installed.
Press enter to continue
```

Same `GHCup-00130` error from the command line.

History: I had a previous installation of ghcup from a few years back that I tried to upgrade and it had this problem.  I ran `ghcup nuke`, deleted `c:\ghc`, and reinstalled from the powershell script, and the same problem is still happening.

dfordivam:
The behavior of existing parser was to intrepret the version as "ghcup" for the path "c:/ghc/ghcup/ghc/<ver>/bin/ghc"
Fixes #1215

### DIFF
--- a/lib/GHCup/Prelude/MegaParsec.hs
+++ b/lib/GHCup/Prelude/MegaParsec.hs
@@ -142,21 +142,11 @@ isSpace c = (c == ' ') || ('\t' <= c && c <= '\r')
 -- ../ghc/<ver>/bin/ghc-<ver>
 ghcVersionFromPath :: MP.Parsec Void Text GHCTargetVersion
 ghcVersionFromPath =
-  (do
-     beforeBin <- parseUntil1 binDir
-     rest  <- MP.getInput
+  do
+     beforeBin <- parseUntil1 binDir <* MP.some pathSep
      MP.setInput beforeBin
      _ <- parseTillLastPathSep
-     x <- ghcTargetVerP
-     MP.setInput rest
-     pure x
-   )
-  <* MP.some pathSep
-  <* MP.takeRest
-  <* MP.eof
+     ghcTargetVerP
   where
      binDir = MP.some pathSep <* MP.chunk "bin" *> MP.some pathSep <* MP.chunk "ghc"
-     parseTillLastPathSep = MP.manyTill MP.anySingle (MP.try $ do
-       _ <- MP.some pathSep
-       _ <- MP.lookAhead (MP.many MP.anySingle)
-       MP.notFollowedBy (MP.manyTill MP.anySingle (MP.some pathSep)))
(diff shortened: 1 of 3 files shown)


lib/GHCup/Prelude/MegaParsec.hs
```diff
@@ -137,3 +137,26 @@ isSpace :: Char -> Bool
...
+   )
+  <* MP.some pathSep
+  <* MP.takeRest
+  <* MP.eof
+  where
+     binDir = MP.some pathSep <* MP.chunk "bin" *> MP.some pathSep <* MP.chunk "ghc"
+     parseTillLastPathSep = MP.manyTill MP.anySingle (MP.try $ do
+       _ <- MP.some pathSep
+       _ <- MP.lookAhead (MP.many MP.anySingle)
+       MP.notFollowedBy (MP.manyTill MP.anySingle (MP.some pathSep)))
```
hasufell, line 162: I don't understand this.

Especially since `MP.lookAhead (MP.many MP.anySingle)` always parses until end of input.

hasufell: I think we can just do:

```hs
     parseTillLastPathSep = (MP.try (parseUntil1 pathSep *> many pathSep) *> p) <|> many MP.anySingle
```

hasufell:
```hs
ghcVersionFromPath :: MP.Parsec Void Text GHCTargetVersion
ghcVersionFromPath =
  do
     beforeBin <- parseUntil1 binDir <* MP.some pathSep
     MP.setInput beforeBin
     _ <- parseTillLastPathSep
     ghcTargetVerP
  where
     binDir = MP.some pathSep <* MP.chunk "bin" *> MP.some pathSep <* MP.chunk "ghc"
     parseTillLastPathSep = (MP.try (parseUntil1 pathSep *> MP.some pathSep) *> parseTillLastPathSep) <|> pure ""
```

This is much cleaner.

The `pure ""` in `parseTillLastPathSep` is important so we always break out of the parser in the first failure (innermost recursion)... if we would parser the ghc version right there and it's a borked ghc version, it would try to re-parse for every filepath component

dfordivam: Ok, yes, recursive is much cleaner.

PR #1244, opened 2025-03-23
Fix logic in guessing incomplete PVP versions, fixes #1243

### LINKED ISSUE #1243: `ghcup set ghc 9.12.1` fails with wrong version
I have ghc 9.12.1 installed through `ghcup`.  However, when I try to set it as the active version, I get this error:
```
$ ghcup set ghc 9.12.1
[ Warn  ] Assuming you meant version 9.12.1.20250219
[ Error ] [GHCup-00130] The version '9.12.1.20250219' of the tool ghc is not installed.
```

hasufell:


lib-opt/GHCup/OptParse.hs
```diff
@@ -186,6 +187,7 @@ opts =
           ))
     <*> invertableSwitch "stack-setup" Nothing False (help "Use stack's setup info for discovering and installing GHC versions")
     <*> (invertableSwitch "paginate" Nothing False (help "Send output (e.g. from 'ghcup list') through pager (default: disabled)"))
+    <*> invertableSwitch "guess-version" Nothing True (help "Whether to guess the full version from an incomplete tool version, e.g. GHC '9.12' resolving to '9.12.2'")
```
hasufell, line 190: This switch is new now... just so that we have a way to have strict matching.

lib-opt/GHCup/OptParse/Compile.hs
```diff
@@ -674,3 +674,6 @@ compile compileCommand settings Dirs{..} runAppState runLogger = do
               VLeft e -> do
                 runLogger $ logError $ T.pack $ prettyHFError e
                 pure $ ExitFailure 9
+ where
+  guessMode = if guessVersion settings then GLaxWithInstalled else GStrict
```
hasufell, line 678: Note that some subcommands set this to `GLaxWithInstalled` and some to `GLax`. E.g. we don't want to match on installed versions during `ghcup install ghc`, but we do during `ghcup set ghc`.

lib/GHCup/Utils/Parsers.hs
```diff
@@ -379,9 +360,40 @@ fromVersion' SetNext tool = do
...
+          ghcs <- if guessMode == GLaxWithInstalled then fmap rights getInstalledGHCs else pure []
+          if v `notElem` ghcs
+          then getLatestToolFor tool (_tvTarget v) pvpIn dls >>= \case
+                 Just (pvp_, vi', mt) -> do
+                   v' <- pvpToVersion pvp_ ""
+                   when (v' /= _tvVersion v) $ logWarn ("Assuming you meant version " <> prettyVer v')
+                   pure (GHCTargetVersion mt v', Just vi')
+                 Nothing -> pure (v, vi)
+          else pure (v, vi)
+    _ -> pure (v, vi)
```
hasufell, line 395: This is the meat of the logic.

hasufell (commented):
The fix is as follows:

* we check if the input version has a full match in the metadata, before starting to guess
  * for some commands, we also check if there is a full match in the *installed* versions
* we add a `--no-guess-version` invertible flag, just in case someone prefers errors in e.g. CI on partial versions

lib/GHCup/Utils/Parsers.hs
```diff
@@ -379,9 +360,40 @@ fromVersion' SetNext tool = do
...
+                 -> GuessMode
+                 -> m (GHCTargetVersion, Maybe VersionInfo)
+guessFullVersion v tool guessMode = do
+  GHCupInfo { _ghcupDownloads = dls } <- getGHCupInfo
+  let vi = getVersionInfo v tool dls
+  case pvp $ prettyVer (_tvVersion v) of -- need to be strict here
+    Left _ -> pure (v, vi)
+    Right pvpIn
+      | (guessMode /= GStrict) && hasn't (ix tool % ix v) dls -> do
+          ghcs <- if guessMode == GLaxWithInstalled then fmap rights getInstalledGHCs else pure []
```
dfordivam, line 386: need to do `getInstalledCabals` etc here?

hasufell: Done

PR #1253, opened 2025-04-13
TUI: allow responding to prompts with Return (defaulting to Yes)

jhrcek:
There's this convention in linux CLI tools that if you're asking users for yes/no response, then you highlight the default choice with uppercase (e.g. [Y/n]) and let user stick with the default by just pressing Enter/Return.

This behavior is already present in bunch of places in ghcup installer. Could we please also support this in the prompts used by TUI?


dfordivam (approved):
LGTM

hasufell:
terrific!

Issue #319, opened 2022-02-01
Mysterious segmentation fault installing ghc 7.10.3


hasufell:
In GitLab by @andreabedini on Feb 1, 2022, 20:43

Hi there,

I am loving ghcup but I just ran into this on my Ubuntu 21.10. Any idea what can cause it or how I can troubleshoot it further?

```
[~] ghcup install ghc 7.10.3
[ Info  ] downloading: https://downloads.haskell.org/~ghc/7.10.3/ghc-7.10.3-x86_64-deb8-linux.tar.bz2 as file /tmp/ghcup-23d985e6f421568b/ghc-7.10.3-x86_64-deb8-linux.tar.bz2
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  133M  100  133M    0     0  3057k      0  0:00:44  0:00:44 --:--:-- 3068k
[ Info  ] verifying digest of: ghc-7.10.3-x86_64-deb8-linux.tar.bz2
[ Info  ] Unpacking: ghc-7.10.3-x86_64-deb8-linux.tar.bz2 to /tmp/ghcup-a338226b2f4807c4
[ Info  ] Installing GHC (this may take a while)
[ ghc-make ] for i in  rts/dist/build/libHSrts.a rts/dist/build/libHSrts_p.a rts/dist/build/libHSrts-ghc7.10.3.so...
[ ghc-make ] "utils/ghc-cabal/dist-install/build/tmp/ghc-cabal-bindist" copy libraries/ghc-prim dist-install "str...
[ ghc-make ] Installing library in
[ ghc-make ] /home/andrea/.ghcup/ghc/7.10.3/lib/ghc-7.10.3/ghcpr_8TmvWUcS1U1IKHT0levwg3
[ ghc-make ] gmake[1]: *** [ghc.mk:924: install_packages] Segmentation fault (core dumped)
[ ghc-make ] gmake: *** [Makefile:24: install] Error 2
[ Error ] Process "gmake" with arguments ["install"] failed with exit code 2.
[ Error ] Also check the logs in /home/andrea/.ghcup/logs
```

EDIT: I can reproduce in a clean Ubuntu 21.10 container.

EDIT2: Reproducible "one-liner"

```
[...]

PR #1254, opened 2025-04-19
Fix issue #319

hasufell:
@mniip

### DIFF
--- a/lib/GHCup/GHC.hs
+++ b/lib/GHCup/GHC.hs
@@ -463,10 +463,21 @@ installUnpackedGHC path inst tver forceInstall addConfArgs installTargets
                        "ghc-configure"
                        Nothing
       tmpInstallDest <- lift withGHCupTmpDir
-      lEM $ make (["DESTDIR=" <> fromGHCupPath tmpInstallDest] <> (words . T.unpack $ installTargets)) (Just $ fromGHCupPath path)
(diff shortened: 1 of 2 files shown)


lib/GHCup/GHC.hs
```diff
@@ -463,10 +463,23 @@ installUnpackedGHC path inst tver forceInstall addConfArgs installTargets
...
       pure ()
+ where
+  installTargets' = T.words installTargets
+
+  -- https://github.com/haskell/ghcup-hs/issues/319
+  withStripTarget ::  Version -> Platform -> [String]
+  withStripTarget ver plat
+    | ver >= [vver|7.10.0|]
+    , ver <  [vver|8.3.0|]
+    , plat `elem` (Linux <$> allDistros)
```
mniip, line 478: `Linux _ <- plat` ?

lib/GHCup/GHC.hs
```diff
@@ -463,10 +463,23 @@ installUnpackedGHC path inst tver forceInstall addConfArgs installTargets
...
+    , not (any (`T.isPrefixOf` "STRIP_CMD=") installTargets')
```
mniip, line 479: This check isn't necessary because in `make VAR=x VAR=y` the `y` will override the `x`.

Issue #1217, opened 2025-01-18
Remove existential types in TUI


hasufell:
https://github.com/haskell/ghcup-hs/blob/d94ce2ae8cb4751c4d6baba7897fd3e1e5390e85/lib-tui/GHCup/Brick/Widgets/Menu.hs#L111-L153

These types must go. It's impossible to derive or write a lens for `fieldInput`.

dfordivam:
Ideally the need for lenses should go away, along with the existential types.(`menuFields    :: [MenuField s n]` must be removed.)

Other good to have things in the advanced install/compile menus.
* show user the version which will be chosen by default. The value of the version field should be initialized to this default value
* a clear button to reset the fields to default values.
* support for mouse clicks
* Allow custom design / placement of buttons etc. ( Clear button should be at the bottom, submit at the top.)

lsmor:
TLDR; Removing existentials would require a whole rewrite of the menus, because current implementation models a menu as a Form, i.e. a list of fields. That list must be heterogenous if we want to pair each field in the form with a field in a record type. Other desing options is having two separated structures and match manually input with field in a record. This implies having a bunch of logic for each menu and lack of re-usability

Just a brief understanding of this. That module is a copy/adaptation of Brick's Form widget, but with optics instead of microlenses (Brick uses microlenses). If I remmeber correctly, there was a good reason to do so. I am talking from memory but I thing the argument goes like this

The general argument goes like this
- we need the user to input many values for some functionalities (like GHC compilation)
- The most usual framework to do this is via a Form
- hence, we draw a form in the tui, gather the inputs, validate per-field (and globally too), and translate into some struct

There are two obvious ways of implementing this. 
1. Using Brick's built-in form or something similar
2. Using a custom form-like logic for each form

I decided to go with the first one because the idea was to have many menus, so we need a way to write those manus easily. The only problem was that Brick uses microlenses, so we needed to add it as a dependency, something undesired back then. Hence, I went for the adaptation of 1. but in reallity we are pretty much doing the same as Brick's Forms. 

Let me first summarize how option two would look alike and why it is a bad option IMO, and later explain the existential types

For option two (custom form-like logic) we need to follow the same workflow "take input" -> "validate" -> "build record" so

```haskell
-- made up types for the sake of example
data GHCCompilationOptions   = {gchversionOption :: ToolVersion,  isolatedpathOption :: Path, ...} -- struct to hold the options
[...]

PR #1255, opened 2025-04-21
TUI rewrite

dfordivam:
- Removes use of existentials. fixes https://github.com/haskell/ghcup-hs/issues/1217

- Makes use of GHC.Generics to take care of the Menu boilerplate

- Improves the structure of code as each widget now need to implement the instance of BaseWidget, which captures the two essential parts, rendering and event handling.

- The structure of the widget state matches directly to the "tree" structure of widgets. And the use of enums, like `Mode`, have been removed.

- The input fields are more generic and we can nest them, like `EditInput` is being used inside `SelectInput`

- The value of input fields can be modified like this

```
  compileGHCMenu % innerWidget % GenericMenu.fields % CompileGHCMenu.installTargetsF % EditInput.editInputOverlay % innerWidget % EditInput.editor %~ Edit.applyEdit (const (textZipper ["new-value"] Nothing))
```

- The child widget does not modify the state of parent (as it was earlier being done via lens)

- The 'quit' key handling logic is now mostly part of the `BasicOverlay` abstraction, and has been removed from the event handlers of individual widgets.

- We can even do nesting of `GenericMenu`, in case we want to make fancy sub-menus


lsmor (commented):
I am not able to fully review this as the code is way more complex than my knowledge :sweat_smile: . My understading is:

- Menus are defined by two parallel datatypes The "Fields" + "Options", and then this `GenericMenu` thing which I barely understand... Instead of Having "Options" + list of fields as before.
- Existentials are hidden behind TypeClasses  + Data.Some if I understand correctly.
- I don't get the Overlays part. AFAIU, `brick` dispatches events to the frontward layers, so there is no need  to pass events to the "child"/overlayed widget... but maybe I got this wrong

dfordivam:
@lsmor Unfortunately the changes here did become big and complex. They can be seen as two major parts

The first is the `BaseWidget` class which describes the overlay functionality, including the passing of events to the widget which is currently open. (`brick` unfortunately has no concept of passing events to frontward layers)
The second is the `GenericMenu` which uses `GHC.Generic`, and does not make use of list of fields with existentials.

I guess the main thing to review here is how the new design is being used and how more features/widgets could be added.

Here I have implemented a confirmation prompt for pre-install messages using this new design. Implementing this in the earlier design would have been far more complex.
 https://github.com/haskell/ghcup-hs/commit/6be847e00e7dd409456988939acf62e20bcd875a

PR #1291, opened 2025-09-16
Fix #1289

### LINKED ISSUE #1289: `ghcup tui` colour for latest version, when uninstalled, is incorrect and may be near-invisible
In `ghcup tui` the tool name and version of of the one marked 'latest', if not installed, is in a different colour from all other tool names and versions: light grey, instead of terminal foreground. This appears to be an error: all other tool names and versions are always in foreground colour, including the one marked 'latest' when it is installed.

Beyond the inconsistency, this causes issues on terminals using dark text on a light background, including `xterm` with its default colours: the name and version may be unreadable or near-unreadable, as in the example below. (See 'HLS 2.11.0.0' and 'cabal 3.16.0.0'.)

Image

   
From comparing the dark-background version with the light-background version, the intent appears to be that all tool names/versions should be in terminal foreground colour. I can't see any particular reason why the tool name/version for just the latest version, and only when uninstalled, should be different.

hasufell:


0cjs:
What does this now look like on light and dark backgrounds?

dfordivam (approved):
LGTM

PR #1304, opened 2026-01-15
Avoid using Alternative unnecessarily

tomjaguarpaw:
Some were completely redundant, some were just unnecessary with a bit of restructuring. This is also a step towards the TODO in `getLinuxDistro`: "don't do alternative on IO, because it hides bugs".

### DIFF
--- a/lib/GHCup/Platform.hs
+++ b/lib/GHCup/Platform.hs
@@ -71,7 +71,7 @@ import qualified Data.List                     as L
 
 
 -- | Get the full platform request, consisting of architecture, distro, ...
-platformRequest :: (MonadReader env m, Alternative m, MonadFail m, HasLog env, MonadCatch m, MonadIO m)
+platformRequest :: (MonadReader env m, MonadFail m, HasLog env, MonadCatch m, MonadIO m)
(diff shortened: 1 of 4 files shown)


lib/GHCup/Platform.hs
```diff
@@ -136,15 +136,15 @@ getPlatform = do
...
 
 
-getLinuxDistro :: (Alternative m, MonadCatch m, MonadIO m, MonadFail m)
+getLinuxDistro :: (MonadCatch m, MonadIO m, MonadFail m)
                => Excepts '[DistroNotFound] m (LinuxDistro, Maybe Versioning)
 getLinuxDistro = do
   -- TODO: don't do alternative on IO, because it hides bugs
-  (name, mid, ver) <- handleIO (\_ -> throwE DistroNotFound) $ lift $ asum
-    [ liftIO try_os_release
+  (name, mid, ver) <- join $ liftIO $ handleIO (\_ -> pure (throwE DistroNotFound)) $ fmap pure $ asum
```
hasufell: Well yeah... the interesting part is getting rid of `asum`.

PR #1364, opened 2026-06-13
Add VSCode integration test

hasufell:


fendor (approved):
LGTM, thank you for the quick action!

Perhaps also add the prerelease channels or something s.t. the issue https://github.com/haskell/vscode-haskell/issues/1374 would have been found as well

fendor:
Yes, we only have very rudimentary integration tests, and we are testing the fallback behaviour for old ghc versions.

hasufell:
@fendor do we give up?

fendor:
No, I was on holiday the last weeks, but I plan to fix this together with the next HLS release.

PR #1370, opened 2026-06-17
Report proper tool name when using --isolate option

marinelli:
At the moment, every time we use `--isolate` option, we have log messages like those ones:
```shell
$ ghcup install cabal --isolate /opt/ghc/bin --force 3.16.1.0
[ Info  ] isolated installing Cabal to /opt/ghc/bin
[ Info  ] cabal installation successful
$ ghcup install ghc --isolate /opt/ghc --force 9.12.4
[ Info  ] isolated installing Cabal to /opt/ghc
[ Info  ] ghc installation successful
```

I suppose it would make sense to use the message `isolated installing` followed by the tool name.


hasufell:
thanks

PR #1379, opened 2026-08-08
Make up/down mappings wrap

Aster89:

Created this as a draft as probably we want to actually make a change to [`brick`](https://hackage-content.haskell.org/package/brick-2.12/docs/Brick-Widgets-List.html) to facilitate and simplify this work.

I've created this PR already because I don't want to lose track of what I'm doing.


hseg:
The PR bumps the minimal `brick` version, but is missing corresponding edits to `stack.yaml` to give a working build plan. The minimal version picked only exists since 2026-08-11, meaning there's only a nightly containing that version.

Aster89:
@hseg, sorry :)

@hasufell, how do I fix that? Do I have to add a line to this?

https://github.com/haskell/ghcup-hs/blob/261a0b0806b8c75ae4ecae16d735c39ccff2803e/stack.yaml#L7-L16

hseg:
Yeah - add the appropriate version of brick to extra-deps. Trying to stack build will give you the precise line to add.

